PR #3, opened 2017-02-24
Multiarch design

twaugh:
This pull request adds a section describing the design for multi-platform container image builds.

@csom, @lcarva

### DIFF
--- a/docs/multiarch.rst
+++ b/docs/multiarch.rst
@@ -1,0 +1,1110 @@
+Multi-platform container image builds
+=====================================
+
+This is the "future state" description of how container images will be
+built for multiple platforms.
+
+A single git commit will be used as the source for building multiple
+container images, each suitable for a different platform (CPU
+architecture, variant, required CPU features, etc).
+
+To support multi-arch builds, two types of OpenShift Build will be
+used:
+
+- *worker builds* which will perform the "docker build" step as
+  before (on compute nodes of the appropriate architecture)
+- *orchestrator build* which will create and manage the worker builds
+
+The cluster which will run orchestrator builds is referred to here as
+the *orchestrator cluster*, and the cluster which will run worker
+builds is referred to here as the *worker cluster*.
+
+Note that the orchestrator cluster itself may also be configured to
+accept worker builds, so the cluster may be both orchestrator and
+worker. Alternatively some sites may want to have complete clusters
+for each platform as well as for the orchestration.
+
+Technically, only compute nodes in the required architectures are
+needed to perform the "docker build" step. These can be arranged
+either in a single mixed-architecture cluster, or with multiple
+single-architecture clusters for each architecture, or as a mix of the
+two.
+
+Client configuration on the Koji builder will decide which cluster
+and/or `node selector`_ is required to perform image builds for each
+required platform.
+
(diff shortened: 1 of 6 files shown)


twaugh:
I've updated this to use graphviz format for diagram sources.

docs/multiarch.rst
```diff
@@ -0,0 +1,1105 @@
...
+- Make alterations to the Dockerfile, including transforming the
+  "FROM" image name and name label into a platform-specific name, as
+  well as setting the ``architecture`` label appropriately
+- "docker build"
+- Squash the image
+- Compress the docker image tar archive
+- Upload this compressed tar archive to Pulp (if Pulp integration and
+  v1 support are both enabled)
+- Tag and push the image to the docker registry, using a unique tag
+  including the timestamp, ``latest``, $version, $version-$release,
```
csomh: Marking all labels as `code` would be nice.

docs/multiarch.rst
```diff
@@ -0,0 +1,1105 @@
...
+
+Tagging
+-------
+
+There are no changes to the tagging scheme. Each image manifest will
+be tagged as before using:
+
+- ``latest``
+- $version
+- $version-$release
```
csomh: ... same here...

docs/multiarch.rst
```diff
@@ -0,0 +1,1105 @@
...
+- any additional tags configured in the git repository
+
+Scratch builds
+--------------
+
+There are no changes to how scratch builds are performed, only some
+parts of the implementation will move around. Some build steps will be
+omitted when performing scratch builds:
+
+- only unique tags will be applied in the worker builds
```
csomh: " only the unique tag including the timestamp will be applied in the worker builds "

docs/multiarch.rst
```diff
@@ -0,0 +1,1105 @@
...
+-------------------
+
+When atomic-reactor in the orchestrator build runs its
+``orchestrate_build`` plugin and watches the builds, it will stream logs
+from those builds and log them again itself with a special prefix
+indicating they are logs relating to a particular platform.
+
+Note that there will be a single Koji task with multiple log
+outputs. When watching this using ``koji watch-logs <task id>`` the log
+output from each worker build will be interleaved. To watch logs from
```
csomh: This might not be very useful most of the time. How is this currently working in koji?

A more useful approach would be that in cases when multiple builds are running for the same task, `koji watch-logs <task id> should only list the available log file names, so that the user can easily pick one of them to watch.

docs/multiarch.rst
twaugh: Like `$version`? These are actually labels not environment variables, and I should probably make that clearer.

docs/multiarch.rst
twaugh: Koji doesn't need changes to work the way it is described here -- it already works this way.

twaugh: Note: for RPMs, binary RPMs for each architecture are built in separate sub-tasks, meaning you can `koji watch-logs ...` the particular task ID you are interested in. For OSBS it's all contained in a single task so that isn't possible.

twaugh:
Updated.

If `--arches <unsupported platform>` is used, the build will fail because there are no clusters available to build for that platform.

csomh:
LGTM

PR #4, opened 2017-03-02
Only allow --arches for scratch builds

twaugh:
Reported by dgilmore.


twaugh:
@csomh, @lcarva

lcarva:
Why do `--arches` param require `--scratch`? Is this to keep parity with how rpms are built?

twaugh:
It is.

PR #5, opened 2017-03-14
Define more orchestrator build metadata

twaugh:
The koji-build-id label and repositories annotations are now defined.


twaugh:
@csomh @lcarva

twaugh:
I had to fix the indentation of that block which is why it looks like a larger change than it is.

docs/multiarch.rst
```diff
@@ -711,56 +711,62 @@ This mechanism can also be used to temporarily disable a worker
...
-            ...
-          durations:
-            koji: ...
-            ...
-          errors: {}
-    plugins-metadata: '{"timestamps": {"orchestrate_build": "...", ...},
-      "durations": {"orchestrate_build": ..., ...}, "errors": {}}'
+    labels:
+      koji-build-id: ...
+    annotations:
```
lcarva: Should we represent `repositories` here as well?

twaugh: Updated to include that. While doing so, I noticed we also want `digests` (for sites without Koji integration).

twaugh:
Updated to avoid duplicating information.

docs/multiarch.rst
```diff
@@ -711,56 +711,82 @@ This mechanism can also be used to temporarily disable a worker
...
+              koji: ...
+              ...
+            durations:
+              koji: ...
+              ...
+            errors: {}
+      plugins-metadata: '{"timestamps": {"orchestrate_build": "...", ...},
+        "durations": {"orchestrate_build": ..., ...}, "errors": {}}'
+
+The existing koji-build-id label is an integer representing the
```
lcarva: It looks like it's converted to a `str`.

twaugh: Fixed.

PR #6, opened 2017-05-02
Update multiarch design with grouping methods

twaugh:
New 'grouping method' concept to deal with both possibilities:
- split content into architecture-specific repositories
- use manifest list to group content in a single repository


twaugh:
I'll merge this as-is, but will submit a new PR to remove the grouping parts.

PR #13, opened 2017-06-27
Describe changes required for logging

twaugh:

### DIFF
--- a/docs/multiarch.rst
+++ b/docs/multiarch.rst
@@ -324,6 +324,150 @@ platforms will continue. Once all have either succeeded or failed, the
 orchestrator build will fail. No content will be available from the
 registry.
 
+.. _`Logging`:
+
+Logging
+-------
+
+Logs from worker builds will be made available via the orchestrator
+build, and clients (including koji-containerbuild) will be able to
+separate individual worker build logs out from that log stream using
+a new osbs-client API method.
+
+Multiplexing
+~~~~~~~~~~~~
+
+In order to allow the client to de-multiplex logs containing a mixture
+of logs from an orchestrator build and from its worker builds, a new
+logging field, platform, is used. Within atomic-reactor all logging
+should be done through a LoggerAdapter which adds this 'platform'
+keyword to the 'extra' dict passed into logging calls. These objects
+should come from a factory function::
+
+  def get_logger(name, platform=None):
+      return logging.LoggerAdapter(logging.getLogger(name),
+                                   {'platform': platform or '-'})
+
+The logging format will include this new field::
+
+  %(asctime)s platform:%(platform)s - %(name)s - %(levelname)s - %(message)s
+
+resulting in log output like this::
+
+  2017-06-23 17:18:41,791 platform:- - atomic_reactor.foo - DEBUG - this is from the orchestrator build
+  2017-06-23 17:18:41,791 platform:x86_64 - atomic_reactor.foo - INFO - 2017-06-23 17:18:41,400 platform:- atomic_reactor.foo -  DEBUG - this is from a worker build
+  2017-06-23 17:18:41,791 platform:x86_64 - atomic_reactor.foo - INFO - continuation line
+
+Demultiplexing will be possible using a new osbs-client API method,
+get_orchestrator_build_logs(). This method is a generator function
+that returns objects with these attributes:
+
+platform
+  str, platform name if worker build, else None
+
+line
+  str, log line (Unicode)
+
+See the example below for what this would look like for these sample
+log lines.
+
+In order to do this it should find the third space-separated field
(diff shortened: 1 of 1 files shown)


docs/multiarch.rst
```diff
@@ -324,6 +324,122 @@ platforms will continue. Once all have either succeeded or failed, the
...
+                                   {'platform': platform or '-'})
+
+The logging format will include this new field::
+
+  %(asctime)s platform:%(platform)s - %(name)s - %(levelname)s - %(message)s
+
+resulting in log output like this::
+
+  2017-06-23 17:18:41,791 platform:- - atomic_reactor.foo - DEBUG - this is from the orchestrator build
+  2017-06-23 17:18:41,791 platform:x86_64 - atomic_reactor.foo -  DEBUG - text:2017-06-23 17:18:41,400 platform:- atomic_reactor.foo -  DEBUG - this is from a worker build
```
lcarva: What's the expected parsing for this line? This:
```
platform = 'x86_64'
line = '2017-06-23 17:18:41,791 platform:x86_64 - atomic_reactor.foo -  DEBUG - text:2017-06-23 17:18:41,400 platform:- atomic_reactor.foo -  DEBUG - this is from a worker build'
```

or this:
```
platform = 'x86_64'
line = '2017-06-23 17:18:41,400 platform:- atomic_reactor.foo -  DEBUG - this is from a worker build'
```
The second form seems more desirable, but not sure how it can be reliably done given possible encoding issues.


Also, where does the `'text:` `str` come from? I don't see it in the formatter

docs/multiarch.rst
```diff
@@ -324,6 +324,122 @@ platforms will continue. Once all have either succeeded or failed, the
...
+- the builder image running atomic-reactor for a worker build
+- the builder image running atomic-reactor for an orchestrator build
+
+When retrieving logs from a build, OpenShift cannot say which encoding
+was used. However, atomic-reactor can define its own output encoding
+to be UTF-8. By doing this, all its log output will be in a known
+encoding, allowing osbs-client to decode it.
+
+In this way, the osbs-client get_build_logs() method will once again
+return an iterable of decoded strings, rather than of a bytes type.
```
lcarva: To do this safely across our toolchain, do we need some kind of versioning for this api method?
I'm thinking that once this is done, koji-containerbuild will need to be updated to properly handle it.

docs/multiarch.rst
```diff
@@ -324,6 +324,122 @@ platforms will continue. Once all have either succeeded or failed, the
...
+
+Orchestrator builds want to retrieve logs from worker builds, then
```
lcarva: Oh! Do we need to set this in our buildroot:
```
ENV LANG en_US.utf8
```

docs/multiarch.rst
lcarva: I see the parsing portion of my question is answered below.

docs/multiarch.rst
twaugh: That (or maybe use LC_ALL), and also have atomic-reactor call setlocale().

docs/multiarch.rst
twaugh: Updated this to describe it better.

Once this is done, koji-containerbuild will use get_orchestrator_build_logs() and de-multiplex the log lines into different files.

docs/multiarch.rst
twaugh: Still unclear about this? See update.

docs/multiarch.rst
```diff
@@ -324,6 +324,143 @@ platforms will continue. Once all have either succeeded or failed, the
...
+If the platform field matches "platform:-", then this is a log line
+from the orchestrator build.
+
+Otherwise, the platform name can be found by removing the "platform:"
+prefix.
+
+For orchestrator build logs, the line is returned as-is.
+
+For worker build logs, the worker log line is returned after this
+modification: if the third field of thw worker build log line matches
```
lcarva: Typo: `thw`

docs/multiarch.rst
lcarva: Clearer now. Although I think it might be worth mentioning that for lines that have been proxied from a worker build, the wrapping orchestrator log message is dropped. For instance, with this line:
```
2017-06-23 17:18:41,791 platform:x86_64 - atomic_reactor.foo - INFO - 2017-06-23 17:18:41,400 platform:- atomic_reactor.foo -  DEBUG - this is from a worker build
```
This whole prefix is removed:
```
2017-06-23 17:18:41,791 platform:x86_64 - atomic_reactor.foo - INFO - 
```

Resulting in this:
```
2017-06-23 17:18:41,400 platform:- atomic_reactor.foo -  DEBUG - this is from a worker build
```

twaugh:
Updated.

PR #16, opened 2017-06-28
Describe add_labels_in_dockerfile

twaugh:
This plugin is now run in the orchestrator build.


lcarva:
We need to move the `add_labels_in_df` definition to after `bump_release`.

twaugh:
Updated.

PR #27, opened 2017-07-14
Document the runPolicy we use

twaugh:

### DIFF
--- a/docs/users.rst
+++ b/docs/users.rst
@@ -92,6 +92,19 @@ Writing a Dockerfile
 Understanding the Build Process
 -------------------------------
 
+OSBS creates an OpenShift BuildConfig to store the configuration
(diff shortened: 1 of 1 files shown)


docs/users.rst
```diff
@@ -92,6 +92,16 @@ Writing a Dockerfile
 Understanding the Build Process
 -------------------------------
 
+OSBS creates an OpenShift BuildConfig to store the configuration
+details about how to run atomic-reactor and which git commit to build
+from. If there is already a BuildConfig for the image (and branch), it
+is updated. Afterwards, a Build is instantiated from the BuildConfig.
+
+The default OpenShift BuildConfig runPolicy of "serial" is used,
```
lcarva: Would be great to mention why SerialLatestOnly is not used.

twaugh: Updated.

PR #28, opened 2017-07-17
Document build priority with node selectors

lcarva:


docs/build_priority.rst
```diff
@@ -0,0 +1,100 @@
...
+Node 3  ✗                   ✗                    ✔
+Node 4  ✗                   ✔                    ✔
+Node 5  ✗                   ✔                    ✔
+======  =================== ==================== ================
+
+In this case, *scratch builds* can be scheduled only on **Node 1**; *explicit
+builds* on any node except **Node 3**; and auto builds on any node except **Node
+2**.
+
+Orchestrator vs Worker Builds
```
rcerven: maybe we can somehow make clear that this is just explanation why we've chosen workers only approach ?

lcarva: Reworded this whole section, PTAL.

PR #31, opened 2017-07-28
Describe new platform:... sections

twaugh:
These sections perform two functions:

* set the GOARCH architecture name for each platform, if different from
  the platform name

* allow publishing v1 images on a per-platform basis

### DIFF
--- a/docs/multiarch.rst
+++ b/docs/multiarch.rst
@@ -503,7 +503,8 @@ Client Configuration
 --------------------
 
 The osbs-client configuration file format will be augmented with
-instance-specific field ``node_selector``.
+instance-specific field ``node_selector``, and new sections for
+describing platforms.
 
 Node selector
 ~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


docs/multiarch.rst
```diff
@@ -523,6 +524,51 @@ the API method ``create_prod_build`` specifying which platform a build
 is required for. If no platform is specified, no node selector will be
 used.
 
+Platform description
+~~~~~~~~~~~~~~~~~~~~
+
+When a section name begins "platform:" it is interpreted not as an
```
lcarva: "begins with" ?

docs/multiarch.rst
```diff
@@ -523,6 +524,51 @@ the API method ``create_prod_build`` specifying which platform a build
...
+OSBS instance but as a platform description. The remainder of the
+section name is the platform name being described. The section has the
+following keys:
+
+architecture (optional)
+  the GOARCH for the platform -- the platform name is assumed to be
+  the same as the GOARCH is this is not specified
```
lcarva: "if this" ?

docs/multiarch.rst
```diff
@@ -523,6 +524,51 @@ the API method ``create_prod_build`` specifying which platform a build
 is required for. If no platform is specified, no node selector will be
 used.
 
+Platform description
```
lcarva: Instead of using a prefix to identify platform specific configurations, what if we introduced a new configuration file? `osbs_platforms.yaml` ?

Not sure if this makes the system more complicated overall. Just mentioning it now because we are unlikely to change it later on.

twaugh: Ideally, the osbs configuration file would already be YAML and we'd add a new key alongside 'instances'.

I think it makes sense to have a single configuration file for this. After all, platform names are used elsewhere in this configuration file, so it's useful to have a nearby reference for what each platform is.

docs/multiarch.rst
twaugh: Fixed.

docs/multiarch.rst
twaugh: Fixed.

twaugh:
Updated, along with the new condition that only one platform is allowed to enable v1.

PR #43, opened 2017-10-02
Use exponential back-off when retrying due to conflict

twaugh:

### DIFF
--- a/docs/contributors.rst
+++ b/docs/contributors.rst
@@ -91,8 +91,12 @@ definitely permanent.
 
 In most cases there should be a delay before retrying. The only reason
 to retry immediately is when the request is likely to succeed when
-immediately retried (i.e. the cause is known). An example would be
-retrying a PUT operation which was rejected due to conflicts.
+immediately retried (i.e. the cause is known). However, caution must
+be exercised. For example, if the cause is a conflict (409), a naive approach
+would be to immediately retry the GET+PUT requests. However this could
+race against some other client doing the same thing. A safer approach
+is to use exponential back-off, retrying the GET+PUT requests at each
+interval.
 
 There should be a maximum number of attempts before giving up and
 reporting an exception.


docs/contributors.rst
```diff
@@ -91,8 +91,12 @@ definitely permanent.
 
 In most cases there should be a delay before retrying. The only reason
 to retry immediately is when the request is likely to succeed when
-immediately retried (i.e. the cause is known). An example would be
-retrying a PUT operation which was rejected due to conflicts.
+immediately retried (i.e. the cause is known). However, caution must
+be exercised. For example, if the cause is a conflict (409), a naive approach
+would be to immediately retry the GET+PUT requests. However this could
+race against some other client doing the same thing. A safer approach
+is to use exponential back-off, retrying the GET+PUT requests at each
```
lcarva: Should there be some randomness to it? To avoid collision over and over again?

csomh: If I [read it correctly](https://en.wikipedia.org/wiki/Exponential_backoff), exponential back-off algorithms do have a randomness in them:

> After c collisions, a random number of slot times between 0 and 2c − 1 is chosen. 

Maybe adding a link to a reference describing the algorithm, or describing the algorithm intended would be worthwhile.

rcerven: how can there be collision over and over ? if there are 2 builds, 1 has to win buildconfig update either way, and that is up to openshift to decide

twaugh: @lcarva, apparently there should. @csomh is describing exponential backoff with full jitter: https://www.awsarchitectureblog.com/2015/03/backoff.html

But @rcerven is also right: in practice we are very unlikely to see a second or third 'conflict' response. New build requests just don't happen frequently enough.

csomh: Why should only the frequency of the build requests considered? Freshmaker probably will trigger a lot's of rebuilds at once.

What was the primary motivation for this change? I don't really understand what is the primary problem to be solved with exponential back-off.

lcarva:
@csomh, @twaugh do we still want this? Looks like there are some unanswered questions.

twaugh:
We can probably drop it.

PR #45, opened 2017-10-16
Add design for ODCS integration

lcarva:
Signed-off-by: Luiz Carvalho <lucarval@redhat.com>

### DIFF
--- a/docs/odcs.rst
+++ b/docs/odcs.rst
@@ -1,0 +1,123 @@
+ODCS Integration
+================
+
+When a container build is started, the user usually provides the URL to a yum
+repo that contains the packages to be installed when building the container
+image. Usually, this is a puddle containing signed packages.
+
+In the case a yum repo URL is not provided, and Koji integration is enabled,
+OSBS falls back to the yum repo associated with the build tag for the Koji build
+target (required parameter for Koji container builds). Packages in this yum repo
+are not always signed. This is not ideal as it’s trivial to produce a container
+image that should never be shipped.
+
+Another problem with the current state is that to produce an image with signed
+packages, a parameter is needed (yum repo URL). To complicate matters, this yum
+repo URL will likely be different for future builds. This poses a huge problem
+for OSBS autorebuilds. In this case, OSBS will automatically rebuild layered
+images upon detected changes to parent image. It is necessary that these
+rebuilds pick up the latest, most relevant, signed packages.
+
+The On Demand Compose Service (`odcs`_) project has been created to address
+current limitations in obtaining a yum repo with signed packages. The general
+idea is to not require the user to provide a yum repo url when building
+container images.  Instead, OSBS will use ODCS to generate a yum repo with the
+relevant signed packages based on user configuration.
+
+API Changes
+-----------
+
+The following parameters will be used by ``buildContainer`` task defined by
+``koji-containerbuild`` plugin:
+
+``odcs_signed <bool, default True>`` When set to False, OSBS will provide an
+empty list of ``sigkeys`` to ODCS. An empty list of ``sigkeys`` indicates the
+signature of packages will not be checked causing the compose to potentially
+include unsigned packages. If ODCS integration is not in place, this parameter
+has no effect.
+
+``container.yaml`` Changes
+--------------------------
+
+In git repo, alongside Dockerfile, the ``container.yaml`` file will be enhanced
+to contain a "compose" section. This section will include all the required
+information for requesting a compose to be created by ODCS::
+
+    compose:
+      # Optional, default False. Inherit compose used for building parent image
+      inherit: true
+      # Optional, default "tag".
+      type: tag
+      # Required for "tag" type. Must contain at least one item.
+      packages:
+      - package-1-name # This is a package name, not an NVR.
+      - package-2-name
+      # Optional. Default value is configured in OSBS environment
+      sigkeys:
+      - AB123456
+
+Alternatively, for modules support::
+
+    compose:
+      type: module
+      # Eventually, ODCS will support ":" separator, for now "-" is used:
+      #     https://pagure.io/odcs/issue/98
+      # Required for "module" type. Must contain at least one item.
+      modules:
+      - "module_name1:stream1"
+      - "module_name2:stream1"
+
+Behavior Changes
+----------------
+
+During an OSBS build, if ``container.yaml`` exists **and** ``compose`` key is defined:
+
+- OSBS orchestrator build will request a compose from ODCS based on specified
+  parameters in ``compose`` section. Once compose completes, build resumes and
+  the URL for the generated compose is passed to worker builds via
+  ``yum_repourls`` build parameter. *NOTE: If odcs-signed parameter was set to
+  false, the compose request will ignore the sigkeys provided in
+  container.yaml.*
+
+- If ``compose`` section sets ``inherit`` to ``true``, the ODCS composes used in
+  building the parent image will be re-used. This may require OSBS to request
+  compose to be regenerated if it’s in *remove* state. Once compose is
+  available, build resumes and the URL for the generated compose is passed to
+  worker builds via ``yum_repourls`` build parameter. *NOTE: this can be used in
+  conjunction with the method stated on previous bullet point. In which case,
+  multiple ODCS composes are used.*
+
+- In the Koji build metadata, ``build.extra.image.odcs.compose_ids`` will be a
+  list of each ODCS compose used. Additionally,
+  ``build.extra.image.odcs.signed`` will correspond to the value of
+  ``odcs_signed`` build parameter which defaults to ``True``
+
+Otherwise, ODCS integration is completely bypassed. The previous behavior of
+using Koji yum repo for build tag of build target is used by default. If
+``yum_repourls`` parameter is used, ODCS integration is also bypassed and only
+the given yum-repos are used. In either case, the Koji build metadata will not
+include any of the ``build.extra.image.odcs.*`` keys.
+
+To clarify, by the time the OSBS worker builds are created, if ODCS is to be
+used, a yum repo URL will be available. This will then be passed to worker
+builds via the usual yum_repourls parameter, ensuring all worker builds use the
+same exact compose.
+
+Relevant Projects
+-----------------
+
+- `koji-containerbuild`_
+
+- `atomic-reactor`_
+
+- `osbs-client`_
+
+- `odcs`_
+
+- `koji`_
+
+.. _`koji-containerbuild`: https://github.com/release-engineering/koji-containerbuild
+.. _`atomic-reactor`: https://github.com/projectatomic/atomic-reactor
+.. _`osbs-client`: https://github.com/projectatomic/osbs-client
+.. _`odcs`: https://pagure.io/odcs
+.. _`koji`: https://pagure.io/koji
--- a/docs/design.rst
+++ b/docs/design.rst
@@ -7,3 +7,4 @@ Designs for future changes
     multiarch
     build_priority
     isolated_builds
+    odcs


docs/odcs.rst
```diff
@@ -0,0 +1,120 @@
...
+empty list of ``sigkeys`` to ODCS causing the compose to potentially include
+unsigned packages. If ODCS integration is not in place, this parameter has no
+effect.
+
+``container.yaml`` Changes
+--------------------------
+
+In git repo, alongside Dockerfile, the ``container.yaml`` file will be enhanced
+to contain a "compose" section. This section will include all the required
+information for requesting a compose to be created by ODCS::
```
vrutkovs: Typo - `::`?

docs/odcs.rst
```diff
@@ -0,0 +1,120 @@
+ODCS Integration
+================
+
+When a container build is started, the user usually provides the URL to a yum
+repo that contains the packages to be installed when building the container
+image. Usually, this is a puddle containing signed packages.
+
+In the case a yum repo URL is not provided, OSBS falls back to the yum repo
+associated with the build tag for the Koji build target (required parameter for
```
twaugh: It's actually optional I think, but I'm not sure if that's because Koji integration is optional, i.e. maybe it's required when Koji integration is enabled, not sure.

docs/odcs.rst
```diff
@@ -0,0 +1,120 @@
...
+
+API Changes
+-----------
+
+The following parameters will be used by ``buildContainer`` task defined by
+``koji-containerbuild`` plugin:
+
+``odcs_signed <bool, default True>`` When set to False, OSBS will not provide an
+empty list of ``sigkeys`` to ODCS causing the compose to potentially include
+unsigned packages. If ODCS integration is not in place, this parameter has no
```
twaugh: I'm  having trouble parsing this sentence because of false/not/empty/unsigned.

docs/odcs.rst
twaugh: It's restructured-text for 'next bit is a literal block'.

docs/odcs.rst
```diff
@@ -0,0 +1,120 @@
...
+
+Alternatively, for modules support::
+
+    compose:
+      type: module
+      # Eventually, ODCS will support ":" separator, for now "-" is used:
+      #     https://pagure.io/odcs/issue/98
+      # Multiple sources can be defined by separating them with a space.
+      # Required for "module" type.
+      source: "module_name1:stream1 module_name2:stream1"
```
twaugh: Should this be a string or a YAML list?

docs/odcs.rst
```diff
@@ -0,0 +1,120 @@
...
+  composes are used.*
+
+- In the Koji build metadata, ``build.extra.image.odcs.compose_ids`` will be a
+  list of each ODCS compose used. Additionally,
+  ``build.extra.image.odcs.signed`` will correspond to the value of
+  ``odcs_signed`` build parameter which defaults to ``True``
+
+Otherwise, ODCS integration is completely bypassed. The previous behavior of
+using Koji yum repo for build tag of build target is used by default. If
+``yum-repo`` parameter is used, ODCS integration is also bypassed and only the
```
twaugh: Which parameter is this?

docs/odcs.rst
vrutkovs: Oh, right, I'm way to used to Markdown

docs/odcs.rst
lcarva: It depends from which POV. If you're kicking this off via rpkg, or koji-cli, it's a required parameter. Should I just remove the "required parameter ..." bit?

docs/odcs.rst
lcarva: I was keeping parity with how ODCS defines it. See bottom of page: https://pagure.io/odcs
Think it'll be cleaner to use something like `modules` here as a list of string?

docs/odcs.rst
lcarva: I guess our various tools call this differently. I'll change it to `yum_repourls` to avoid confusion.

lcarva: Done.

docs/odcs.rst
lcarva: Changed it to `modules` list. I'm wondering if `type` now is necessary.

docs/odcs.rst
lcarva: That's because it wasn't written correctly. Fixed.

docs/odcs.rst
lcarva: Reworded this paragraph to indicate this applies when Koji integration is enabled.

docs/odcs.rst
```diff
@@ -0,0 +1,123 @@
...
+``container.yaml`` Changes
+--------------------------
+
+In git repo, alongside Dockerfile, the ``container.yaml`` file will be enhanced
+to contain a "compose" section. This section will include all the required
+information for requesting a compose to be created by ODCS::
+
+    compose:
+      # Optional, default False.
+      inherit: true
```
csomh: `inherit` is nice and short, but it feels not self-explanatory enough, would `inherit_from_parent` work?

docs/odcs.rst
```diff
@@ -0,0 +1,123 @@
...
+      #     https://pagure.io/odcs/issue/98
+      # Required for "module" type. Must contain at least one item.
+      modules:
+      - "module_name1:stream1"
+      - "module_name2:stream1"
+
+Behavior Changes
+----------------
+
+During an OSBS build, if ``container.yaml`` exists **and** ``container`` key is defined:
```
csomh: 'compose' key

docs/odcs.rst
```diff
@@ -0,0 +1,123 @@
...
+
+- OSBS orchestrator build will request a compose from ODCS based on specified
+  parameters in ``container`` section. Once compose completes, build resumes and
```
csomh: 'compose' section

docs/odcs.rst
```diff
@@ -0,0 +1,123 @@
...
+  the URL for the generated compose is passed to worker builds via
+  ``yum_repourls`` build parameter. *NOTE: If odcs-signed parameter was set to
+  false, the compose request will ignore the sigkeys provided in
+  container.yaml.*
+
+- If ``container`` section sets ``inherit`` to ``true``, the ODCS composes used
```
csomh: 'compose' section

lcarva: Thanks! Fixed all those.

docs/odcs.rst
lcarva: I think `inherit_from_parent` is a bit too long and more error prone. I clarified the comment above to say: 
>  Optional, default False. Inherit compose used for building parent image

docs/odcs.rst
```diff
@@ -0,0 +1,123 @@
...
+
+Behavior Changes
+----------------
+
+During an OSBS build, if ``container.yaml`` exists **and** ``compose`` key is defined:
+
+- OSBS orchestrator build will request a compose from ODCS based on specified
+  parameters in ``compose`` section. Once compose completes, build resumes and
+  the URL for the generated compose is passed to worker builds via
+  ``yum_repourls`` build parameter. *NOTE: If odcs-signed parameter was set to
```
owtaylor: Passing via yum_repourls is going to break the Flatpak flow because for Flatpak, the worker builds need to be able to find out what exact module versions ODCS resolved the source specification to, and retrieve the modulemd files for those module versions. That could be worked around by passing the resolved module versions to the worker builds as well, but would be easier to just leave the current system where the compose_id is passed - the extra round-trip to ODCS to retrieve the URL should be a matter of milliseconds. 

As I mentioned above, it may also be interesting as well for higher level tooling to be able to to pass in a compose ID to be able to control the compose process exactly - at least in the modular case.

docs/odcs.rst
```diff
@@ -0,0 +1,123 @@
...
+  false, the compose request will ignore the sigkeys provided in
+  container.yaml.*
+
+- If ``compose`` section sets ``inherit`` to ``true``, the ODCS composes used in
+  building the parent image will be re-used. This may require OSBS to request
+  compose to be regenerated if it’s in *remove* state. Once compose is
+  available, build resumes and the URL for the generated compose is passed to
+  worker builds via ``yum_repourls`` build parameter. *NOTE: this can be used in
+  conjunction with the method stated on previous bullet point. In which case,
+  multiple ODCS composes are used.*
```
owtaylor: *How* can multiple ODCS composes by used?

docs/odcs.rst
```diff
@@ -0,0 +1,123 @@
...
+container images.  Instead, OSBS will use ODCS to generate a yum repo with the
+relevant signed packages based on user configuration.
+
+API Changes
+-----------
+
+The following parameters will be used by ``buildContainer`` task defined by
+``koji-containerbuild`` plugin:
+
+``odcs_signed <bool, default True>`` When set to False, OSBS will provide an
```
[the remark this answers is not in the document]
lcarva: > I think "ODCS" here is a detail and shouldn't in the parameter name.

It's a detail of the implementation, true. But I don't think we can get away with something less specific without causing confusion. If this parameter was called simply `signed`, it's ambiguous that it only applies to when requesting a new compose from ODCS. Perhaps when ODCS has ubiquitous adoption we may want to revisit.



> The behavior with respect to sigkeys in container.yaml should be specified. Is it?
> 
> If False, sigkeys in container.yaml is ignore
> If True, a non-empty list of sigkeys is required, and one of them must be found.

`sigkeys` in `container.yaml` is optional. If not defined, the OSBS environment will define the default value. But the user has the option to change it (use beta sigkey for example). If the user does decide to use different sigkeys, the list must contain at least one item. (maybe this needs to be written out more explicitly in design doc?) 

To make ODCS not care about package signatures, an empty list of `sigkeys` must be provided. So when `odcs_signed` is set to `False`, an empty list of signkeys is always passed when requesting a compose from ODCS. Otherwise, it uses either what's in container.yaml or the default value in OSBS environment.



> Does it really make sense to say "definitely give me a package from a signed compose, but I don't care if it's provisionally signed or finally signed".

If the intent is "provisional", why should the compose fail if one of the packages (could be dependency) is signed with the "final" key?



> My expectation is that the person (or service) running the containerBuild task should provide an intent ("unsigned", "provisional", "public")

I like this approach. Good call on pointing out multiple repos with key IDs checked in. Providing the intent should allow OSBS to better manage which keys are used. I think a mapping, as you suggested, makes perfect sense. For the "unsigned" intent, it'll essentially map to an empty list of sigkeys.

docs/odcs.rst
```diff
@@ -0,0 +1,123 @@
...
+
+``container.yaml`` Changes
+--------------------------
+
+In git repo, alongside Dockerfile, the ``container.yaml`` file will be enhanced
+to contain a "compose" section. This section will include all the required
+information for requesting a compose to be created by ODCS::
+
+    compose:
+      # Optional, default False. Inherit compose used for building parent image
```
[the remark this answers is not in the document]
lcarva: The idea here is that for images that are part of the same product line, for instance, we may want to generate a single compose to be used by the whole set of images. This has two main benefits:
1. If there are multiple in-flight releases, once the top most image is built (compose generated), the layered images won't accidentally pick up unrelated newer packages.
2. It provides an easier transition path from the existing approach of using puddles.

It might be premature to include this feature now. Maybe it should be omitted until there's an actual ask for it.

@twaugh, thoughts?

docs/odcs.rst
```diff
@@ -0,0 +1,123 @@
...
+      inherit: true
+      # Optional, default "tag".
+      type: tag
```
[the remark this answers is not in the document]
lcarva: > How do you actually specify the tag?

Tag is the Koji build tag for the Koji build target.



> Suggest this should be required. Plan is to be much more modular in the future, and then making this default to tag just is going to result in confusing error messages.

I wonder if this can be inferred. If we see `packages`, assume source is tag. If we see `modules`, assume source is module. If we see both or neither, that's an error. What do you think?

docs/odcs.rst
lcarva: Sold! Wasn't aware the worker builds required additional information from compose other than repo file URL.

docs/odcs.rst
lcarva: This goes together with the `inherit` feature. Potentially, each layered image can generate their own compose as well as re-use the ones from parent.

lcarva:
> As discussed in person, this approach seems generally compatible with what is already being done for the Flatpak case. The container.yaml approach could be an alternative to what is currently being done by passing a module through koji-containerbuild / osbs-client or simply replace it entirely.

If it's not too disruptive, I'd rather converge into the same approach.

> The main advantage of the approach of passing a module on the command line is that it gives the opportunity for specifying a particular module version to build and not just a stream - this might be useful for tooling that wants to build a container out of a module version that has already passed CI checks.
> 
> But the caveat to that, and why it doens't matter that much - is that you can only specify the module version for the leaf module - and not the full resolved module stack. So, you can specify a particular version of mariadb, but the compose will end up with whatever is the latest build of the Platform module. For this case , it might be better for the tooling to request the compose itself via ODCS, and pass in a compose ID.

[Freshmaker](https://pagure.io/freshmaker) has a similar use case. They take care of generating whatever repos are needed, then pass in the yum repo urls. I can see how for flatpak that can be challenging, as additional info from the ODCS compose is needed. Probably taking ODCS compose ID should be fine as a build parameter. But we must keep in mind that passing arguments like this is really not ideal. Maybe the new compose ID parameter can only be used with scratch builds?

docs/odcs.rst
owtaylor: So the idea is that you'd do two composes, inject both yum repourls, and hope for the best? I don't think you achieve "the layered images won't accidentally pick up unrelated newer packages" - any newer packages that are dependencies of the packages listed by the layered image (glibc, etc.) will be pulled into the compose.

[ ODCS seems to always invoke pungi with `koji_inherit = false`, which would help in this regard, but doesn't generally seem reasonable  - probably this just has never been tested in a practical case. ]

docs/odcs.rst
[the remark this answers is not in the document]
owtaylor: You are right that the "provisional" case provides a case where a single intent can map to a list of signatures - it's fine to get "final" packages in that case. It's the other way that doesn't work.

docs/odcs.rst
lcarva: > [ ODCS seems to always invoke pungi with koji_inherit = false, which would help in this regard, but doesn't generally seem reasonable - probably this just has never been tested in a practical case. ]

This will soon change to actually be true, which I guess further emphasizes your point :)

owtaylor:
> As discussed in person, this approach seems generally compatible with what is already being done for the Flatpak case. The container.yaml approach could be an alternative to what is currently being done by passing a module through koji-containerbuild / osbs-client or simply replace it entirely.
> 
> If it's not too disruptive, I'd rather converge into the same approach.

It shouldn't be too disruptive - I have a strong goal of trying to keep things the same for server-side containers and Flatpaks, so if we're asking maintainers to learn how to add container.yaml files for server-side containers, I don't mind asking for that for client-side containers as well.

> But we must keep in mind that passing arguments like this is really not ideal. Maybe the new compose ID parameter can only be used with scratch builds?

If passing a compose ID is in useful for Freshmaker, it's definitely not just for scratch builds. The *reason* that passing arguments is not ideal is mostly that we want builds to be repeatable and updateable in an automated fashion. That implies:

 * Permanent state (build against the f26 stream) is ideally not passed to containerBuild, it comes from the repository.
 * State that *is* passed to the containerBuild - should be interpretable, not opaque. It's better to have a a compose ID than a repository URL, it's better to have an "intent" for signing, then a list of signing keys. And it should be preserved in the resulting metadata for the build.

lcarva:
> If passing a compose ID is in useful for Freshmaker, it's definitely not just for scratch builds.

AFAIK, Freshmaker could pass in compose ID, but this would actually make it more complex for them because they pass additional repos for content sets. We'd end up in a situation where we'd have to handle both yum repourls and compose IDs on the same build.



> The reason that passing arguments is not ideal is mostly that we want builds to be repeatable and updateable in an automated fashion.

Yes, also every parameter that is passed, is written to the BuildConfig which is the input for the autorebuilds in OSBS. It can cause quite an amount of confusion if an image is automatically rebuilt with the compose ID of the last build run! Scratch (and isolated) builds don't use a BuildConfig, so they don't affect subsequent builds.

(We're doing something different for `odcs_signed=False` by removing this from BuildConfig immediately after Build object is instantiated. But I'd rather minimize this sort of behavior.)

owtaylor:
> If passing a compose ID is in useful for Freshmaker, it's definitely not just for scratch builds.
> 
> AFAIK, Freshmaker could pass in compose ID, but this would actually make it more complex for them because they pass additional repos for content sets. We'd end up in a situation where we'd have to handle both yum repourls and compose IDs on the same build.

I hope that you have the freshmaker people lined up to review this plan :-)
 
>     The reason that passing arguments is not ideal is mostly that we want builds to be repeatable and updateable in an automated fashion.
> 
> Yes, also every parameter that is passed, is written to the BuildConfig which is the input for the autorebuilds in OSBS. It can cause quite an amount of confusion if an image is automatically rebuilt with the compose ID of the last build run! Scratch (and isolated) builds don't use a BuildConfig, so they don't affect subsequent builds.

Is there a plan that builds from freshmaker will become targets for OSBS rebuilds? Or are they orthogonal and separate rebuilding systems?

> (We're doing something different for odcs_signed=False by removing this from BuildConfig immediately after Build object is instantiated. But I'd rather minimize this sort of behavior.)

Removing odcs_signed=False from the BuildConfig is weird - that means an unsigned build turns into a signed build at the first build. In the world of automatic rebuilds, it seems like signature behavior needs to be consistent policy, not a build option.

lcarva:
> I hope that you have the freshmaker people lined up to review this plan :-)

After talking to the Freshmaker team, I realize now that both yum_repourls they provide map to ODCS composes. One for koji tag source and one for pulp source. We can see the benefit of having the compose ID being passed in instead of the yum_repourls. (OSBS can extend/renew ODCS compose if needed; same tracking mechanism in Koij build metadata).
I'll update the follow up PR, #46, with this.




> Is there a plan that builds from freshmaker will become targets for OSBS rebuilds? Or are they orthogonal and separate rebuilding systems?

Freshmaker uses OSBS for its container rebuilds via [isolated builds](http://osbs.readthedocs.io/en/latest/isolated_builds.html) (no BuildConfig in OpenShift).



> Removing odcs_signed=False from the BuildConfig is weird - that means an unsigned build turns into a signed build at the first build. In the world of automatic rebuilds, it seems like signature behavior needs to be consistent policy, not a build option.

This is precisely what we're trying to achieve. By removing the attribute from the BuildConfig, an automatically created build (parent image changed) will produce an image with signed packages.

PR #46, opened 2017-10-17
Odcs integration - Part 2

lcarva:


lcarva:
cc @owtaylor

docs/odcs.rst
```diff
@@ -46,46 +46,49 @@ information for requesting a compose to be created by ODCS::
...
 During an OSBS build, if ``container.yaml`` exists **and** ``compose`` key is defined:
 
 - OSBS orchestrator build will request a compose from ODCS based on specified
   parameters in ``compose`` section. Once compose completes, build resumes and
-  the URL for the generated compose is passed to worker builds via
-  ``yum_repourls`` build parameter. *NOTE: If odcs-signed parameter was set to
-  false, the compose request will ignore the sigkeys provided in
-  container.yaml.*
+  the compose ID is passed to the worker builds via the new ``compose_ids``
+  build parameter. *NOTE: If odcs-signed parameter was set to false, the compose
```
csomh: hm... did I get lost in the diffs? Where should this parameter be set? Will this be a value in container.yaml?

Or this is meant as `intent` not defined in container.yaml?

lcarva: `odcs-signed` is ultimately a CLI parameter, that gets propagated to koji task, and eventually osbs create build api.

csomh: Gotcha, no questions then.

docs/odcs.rst
```diff
@@ -36,6 +36,10 @@ signature of packages will not be checked causing the compose to potentially
 include unsigned packages. If ODCS integration is not in place, this parameter
 has no effect.
 
+``compose_id [ID [ID ...]]`` List of ODCS compose IDs to be used. When defined,
+``compose`` section of ``container.yaml`` is ignored. Cannot be used with
+``repo-url``, nor ``odcs_signed`` parameters.
```
twaugh: ``yum_repourls``?

docs/odcs.rst
```diff
@@ -46,46 +50,49 @@ information for requesting a compose to be created by ODCS::
...
-      sigkeys:
-      - AB123456
+      # Optional. Default and possible values are configured in OSBS environment
+      # which will be a mapping of intent -> list of signing keys
+      #     release -> my-release-key
+      #     beta -> my-beta-key, my-release-key
+      #     unsigned -> <empty>
+      intent: release
 
 Alternatively, for modules support::
```
twaugh: Not having 'type', and inferring it from the presence of either 'packages' or 'modules', *seems* OK. As long as it can be described by a JSON schema, I'm fine with doing it that way, and I think it can ("oneOf").

docs/odcs.rst
```diff
@@ -46,46 +50,49 @@ information for requesting a compose to be created by ODCS::
...
       - package-2-name
-      # Optional. Default value is configured in OSBS environment
-      sigkeys:
-      - AB123456
+      # Optional. Default and possible values are configured in OSBS environment
+      # which will be a mapping of intent -> list of signing keys
+      #     release -> my-release-key
+      #     beta -> my-beta-key, my-release-key
+      #     unsigned -> <empty>
+      intent: release
```
twaugh: Do we want to use `intent` as the buildContainer parameter instead of `odcs_signed`?

twaugh: Similarly, should the intent appear in the Koji build metadata?

docs/odcs.rst
lcarva: I'm going to clean this up, but yes and no :)
`repo-url` is the CLI option, while `yum_repourls` is the API option.

docs/odcs.rst
lcarva: This probably makes more sense at this point. Do we want `odcs_intent`? Or just `intent`?
Reason being that if ODCS integration is not enabled, `intent` is meaningless.

owtaylor: Or `signing_intent` to make the key more self-documenting?

lcarva: +1 for `signing_intent` as the key in yaml file
Maybe the same for cli/api (`signing-intent/signing_intent`) parameter to avoid confusion?

lcarva:
@twaugh, @owtaylor, @csomh, I made various changes based on feedback:

- Use signing_intent instead of sigkeys directly
- Removal of parent image compose inheritance

Additionally, I added the concepts section to better explain how some of these ideas work together.

The doc is probably better reviewed now in full :(

docs/odcs.rst
lcarva: Added `signing_intent`.

docs/odcs.rst
lcarva: Added API/CLI Changes sections to clarify this.

docs/odcs.rst
```diff
@@ -24,84 +24,188 @@ idea is to not require the user to provide a yum repo url when building
...
+Next, ``container.yaml`` must be created and it must include the *compose*
+section. See example further down. This file must then be committed and pushed
+to git repository. Simply building the container image as usual will trigger
+ODCS interaction.
+
+
+ODCS Compose Checks
+"""""""""""""""""""
+
+When an ODCS compose is passed in as an argument, OSBS will perform some checks
```
csomh: Would it be better to say 'ODCS compose ID'?

docs/odcs.rst
```diff
@@ -24,84 +24,188 @@ idea is to not require the user to provide a yum repo url when building
...
+
+  May not be used if ODCS integration is not enabled.
+
+  May not be used with ``signing_intent`` parameter.
+
+  Ignores *compose* section in ``container.yaml`` and does not request a new
+  ODCS compose to be created. The provided composes are used instead.
+
+
+CLI Changes
```
csomh: Not sure if did understand correctly: will these CLI options override the values from container.yaml?

If so, shouldn't these be restricted to scratch builds only? (so that these values are always have a trace in dist-git for builds which can be potentially released)

docs/odcs.rst
lcarva: Yes. Fixed.

docs/odcs.rst
lcarva: > will these CLI options override the values from container.yaml?

Yes. This is needed to give some flexibility when building a hierarchy of images without signed packages. The pattern I had considered was to have the most restrictive signing intent in `container.yaml`. In most cases, this would be "release", but for upcoming products this could be "beta". The CLI argument would allow users to downgrade this to "unsigned", for example. This is critical in allowing workflows for daily builds. Currently, initiating signing of packages is done via a manual process which is somewhat time consuming.

I hadn't considered the scenario of upgrading the signing intent. For instance, if `container.yaml` defines "beta", but "release" is used in the CLI. I guess in this case it's fine. If build is successful, it really means that packages used have been signed with the "release" intent. This value is also recorded in the Koji build metadata for tracking purposes.

docs/odcs.rst
```diff
@@ -24,84 +24,188 @@ idea is to not require the user to provide a yum repo url when building
...
+
+*This mapping will be defined in reactor-config-secret which will allow updates
+to be reflected instantly for automatically triggered builds from an existing
+BuildConfig.*
+
+The definition of intents also describes the restrictive order, which will be
+enforced when building layered images. For instance, consider the case of two
+images, X and Y. Y uses X as its parent image (FROM X). If image X was built
+with "beta" intent, image Y's intent can only be "beta" or "unsigned". If the
+repo for image Y has it configured to use "release" intent, this value will be
```
twaugh: Meaning the dist-git repository?

lcarva: Correct. Changed it to "dist-git repo" for clarification.

docs/odcs.rst
```diff
@@ -24,84 +24,188 @@ idea is to not require the user to provide a yum repo url when building
...
 API Changes
 -----------
 
-The following parameters will be used by ``buildContainer`` task defined by
-``koji-containerbuild`` plugin:
+- ``signing_intent <str, default None>``
+
+  Exposed in `koji-containerbuild`_, and `osbs-client`_.
+
+  May not be used if ODCS integration is not enabled.
```
twaugh: Is that "May be ignored" or "Must not be used"?

docs/odcs.rst
```diff
@@ -24,84 +24,188 @@ idea is to not require the user to provide a yum repo url when building
...
+
+Signing Intent
+""""""""""""""
+
+When an ODCS compose is used, the idea of a signing intent is used. This refers
+to the signature of the content to be included in the container image. For
+instance, an environment may provide the following intents: ``release``,
+``beta``, and ``unsigned``. Each one of those intents is then mapped to a list
+of signing keys. These signing keys are then used during compose creation. The
+packages to be included must have been signed by either one of the keys listed.
```
twaugh: "either one" reads a little awkwardly there. Maybe "any"?

docs/odcs.rst
```diff
@@ -24,84 +24,188 @@ idea is to not require the user to provide a yum repo url when building
...
+
+    # Only include packages that have been signed by "my-release-key"
+    release -> my-release-key
+    # Include packages that have been signed by either "my-beta-key" or
+    # "my-release-key"
+    beta -> my-beta-key, my-release-key
+    # Do not check signature of packages - may include unsigned packages
+    unsigned -> <empty>
+
+*This mapping will be defined in reactor-config-secret which will allow updates
```
twaugh: Maybe call this "the secret named by reactor_config_secret".

docs/odcs.rst
```diff
@@ -24,84 +24,188 @@ idea is to not require the user to provide a yum repo url when building
...
+to be reflected instantly for automatically triggered builds from an existing
+BuildConfig.*
+
+The definition of intents also describes the restrictive order, which will be
+enforced when building layered images. For instance, consider the case of two
+images, X and Y. Y uses X as its parent image (FROM X). If image X was built
+with "beta" intent, image Y's intent can only be "beta" or "unsigned". If the
+dist-git repo for image Y has it configured to use "release" intent, this value
+will be downgraded to "beta" at build time.
```
twaugh: I'm a bit confused by this. It seems to merge together the intent and the result, in that the developer's intent can be altered by the build system.

Why wouldn't we want this build to fail? Compare with the distribution-scope label enforcement: if the parent image's distribution scope is narrower than the image we're attempting to build, we fail.

I'm sure there is a reason for doing it the way it's documented here, but can we document what that reason is?

Also, how is this build comparing with the parent build's intent? By examining the Koji metadata for that parent build?

docs/odcs.rst
twaugh: I'm also a bit wary of not also requiring scratch builds when setting this values using the CLI. Compare with overriding architectures from the CLI, which cannot be done unless the build is a scratch build. Can we include some rationale on that?

docs/odcs.rst
```diff
@@ -24,84 +24,188 @@ idea is to not require the user to provide a yum repo url when building
...
+  ODCS compose to be created. The provided composes are used instead.
+
+
+CLI Changes
+-----------
+
+These mostly correspond to the API changes above. It's listed here mainly to
+emphasize they may be spelled differently.
+
+- ``signing-intent <str, default None>``
```
twaugh: `--signing-intent`?

docs/odcs.rst
```diff
@@ -24,84 +24,188 @@ idea is to not require the user to provide a yum repo url when building
...
-  multiple ODCS composes are used.*
+  parameters in *compose* section. Once compose completes, build resumes and
+  the compose ID is passed to the worker builds via the new ``compose_ids``
+  build parameter.
 
 - In the Koji build metadata, ``build.extra.image.odcs.compose_ids`` will be a
   list of each ODCS compose used. Additionally,
-  ``build.extra.image.odcs.signed`` will correspond to the value of
-  ``odcs_signed`` build parameter which defaults to ``True``
+  ``build.extra.image.odcs.signed_intent`` will correspond to the signing intent
```
twaugh: Should this be `build.extra.image.odcs.signing_intent`?

docs/odcs.rst
csomh: Would it make sense to have an example for this secret with these changes?

docs/odcs.rst
lcarva: Yes. Fixed.

docs/odcs.rst
lcarva: Yes. Added `--` prefix (also for `compose-id` CLI parameter).

docs/odcs.rst
lcarva: Thanks. Changed it to "Must not be used".

docs/odcs.rst
lcarva: Replaced `reactor-config-secret` with suggested text.

Added example yaml.

docs/odcs.rst
lcarva: Changed it to "any".

docs/odcs.rst
lcarva: Added this:
> The Koji build metadata will contain a new key, build.extra.image.odcs.signing_intent_overridden, to indicate whether or not the signing_intent was overridden (CLI parameter, automatically downgraded, etc). This value will only be true if build.extra.image.odcs.signing_intent matches the signing_intent in container.yaml.

lcarva:
> I think it might be better to have a well-known enumeration of signing intents, rather than just saying it is defined in the site OSBS configuration.

To recap our offline conversation, the well-known enumeration is highly dependable on a given OSBS environment. As such, that should be documented wherever it makes the most sense for the OSBS environment.

docs/odcs.rst
lcarva: > how is this build comparing with the parent build's intent? By examining the Koji metadata for that parent build?

That's correct. It'll use the Koji metadata for the parent build.

Added this paragraph to further clarify the rationale behind this behavior:
> Automatically downgrading the signing intent, instead of failing the build, is important for allowing a hierarchy of layered images to be built automatically by ImageChangeTriggers. For instance, with Continuous Integration in mind, a user may want to perform daily builds without necessarily requiring signed packages while periodically also producing builds with signed content. In this case, the signing_intent in container.yaml can be set to release for all the images in hierarchy. Whether or not the layered images in the hierarchy use signed packages can be controlled by simply overriding the signing intent of the top most image. The signing intent of the layered images would then be automatically adjusted as needed.

lcarva:
@twaugh, I think I have addressed all of your comment. PTAL.

docs/odcs.rst
```diff
@@ -24,84 +24,214 @@ idea is to not require the user to provide a yum repo url when building
...
+
+Automatically downgrading the signing intent, instead of failing the build, is
+important for allowing a hierarchy of layered images to be built automatically
+by ImageChangeTriggers. For instance, with Continuous Integration in mind, a
+user may want to perform daily builds without necessarily requiring signed
+packages while periodically also producing builds with signed content. In this
+case, the ``signing_intent`` in ``container.yaml`` can be set to ``release`` for
+all the images in hierarchy.  Whether or not the layered images in the hierarchy
+use signed packages can be controlled by simply overriding the signing intent of
+the top most image. The signing intent of the layered images would then be
```
twaugh: People imagine layers different ways up depending on how their brains work. Is there a way of wording this that avoids "top"?

docs/odcs.rst
```diff
@@ -24,84 +24,214 @@ idea is to not require the user to provide a yum repo url when building
...
+the build parameter of same name. This particular parameter will be ignored for
+autorebuilds. The value in ``container.yaml`` should always be used in that
+case. Note that the signing intent used by the compose of parent image is still
+enforced.
+
+The Koji build metadata will contain a new key,
+``build.extra.image.odcs.signing_intent_overridden``, to indicate whether or not
+the ``signing_intent`` was overridden (CLI parameter, automatically downgraded,
+etc).  This value will only be ``true`` if
+``build.extra.image.odcs.signing_intent`` matches the ``signing_intent`` in
```
twaugh: Should that be "does not match"?

docs/odcs.rst
```diff
@@ -24,84 +24,214 @@ idea is to not require the user to provide a yum repo url when building
...
+  parameters in *compose* section. Once compose completes, build resumes and
+  the compose ID is passed to the worker builds via the new ``compose_ids``
+  build parameter.
 
 - In the Koji build metadata, ``build.extra.image.odcs.compose_ids`` will be a
   list of each ODCS compose used. Additionally,
-  ``build.extra.image.odcs.signed`` will correspond to the value of
-  ``odcs_signed`` build parameter which defaults to ``True``
+  ``build.extra.image.odcs.signing_intent`` will correspond to the signing
+  intent of the ODCS composes.
```
twaugh: Also mention `build.extra.image.odcs.signing_intent_overridden` here?

docs/odcs.rst
lcarva: Yes, of course. Fixed it.

docs/odcs.rst
lcarva: Good idea. Changed this paragraph to this format:

> In the Koji build metadata:
> - build.extra.image.odcs.compose_ids: list of each ODCS compose used.
> - build.extra.image.odcs.signing_intent: lest restrictive signing intent of the ODCS composes.
> - build.extra.image.odcs.signing_intent_overridden: whether or not the signing intent used is different than the one defined in container.yaml.

docs/odcs.rst
lcarva: I'm rattling my brain on this one :)

How about "top most ancestor image" ? It might better clarify which direction we mean.

docs/odcs.rst
twaugh: Can signing_intent_overridden be absent, and if so should it be assumed to be false?

docs/odcs.rst
twaugh: Yes, that's better.

docs/odcs.rst
```diff
@@ -24,84 +24,220 @@ idea is to not require the user to provide a yum repo url when building
...
-  ``odcs_signed`` build parameter which defaults to ``True``
+  parameters in *compose* section. Once compose completes, build resumes and
+  the compose ID is passed to the worker builds via the new ``compose_ids``
+  build parameter.
+
+- In the Koji build metadata:
+
+    - ``build.extra.image.odcs.compose_ids``: list of each ODCS compose used.
+
+    - ``build.extra.image.odcs.signing_intent``: lest restrictive signing intent
```
twaugh: typo: least

But I wonder if this should be described as 'final signing intent (after taking parent image's signing intent into account)' or something.

docs/odcs.rst
lcarva: Done.

docs/odcs.rst
lcarva: Changed it to:
> final signing intent of the ODCS composes after adjusting for CLI parameter, automatically downgraded, etc.

docs/odcs.rst
lcarva: I'm thinking it's better to always have it in there. Makes consuming this data easier.

PR #53, opened 2017-11-22
Pull in some of the multiarch.rst parts to the main body of the text

twaugh:
This PR folds all the relevant parts of multiarch.rst into the main body of the text in the right places.


twaugh:
Updated, actually now it is hopefully complete.

docs/build_process.rst
```diff
@@ -0,0 +1,362 @@
...
+the registry, grouping together the image manifests from the worker
+builds.
+
+.. _`manifest list`: https://docs.docker.com/registry/spec/manifest-v2-2/#manifest-list
+
+.. graphviz:: images/arrangement-v4.dot
+   :caption: Orchestrator and worker builds (arrangement v4)
+
+Arrangement version 5 (ODCS Integration)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
csomh: I think this should use `---` for title, otherwise becomes a child of arrangement version 4.

docs/admins.rst
```diff
@@ -1,25 +1,122 @@
...
-The /etc/osbs.conf file used by the koji-containerbuild plugin must
-name secrets configured on the OpenShift orchestration cluster:
+- the one external to OSBS, for creating orchestrator builds, and
+- the one internal to OSBS, stored in a Kubernetes secret (named by
+  `client_config_secret`_) in the orchestrator cluster
+
+These can have the same content. The important features are discussed
+below.
+
+Can Orchestrate
```
csomh: I'm wondering if this and `Token secrets` shouldn't be rather written as `can_orchestrate` and `token_secrets`, to match in style with `reactor_config_secret` and `client_config_secret`. (Or a similar change the other way around - reading the TOC feels inconsistent this way).

twaugh: How about 'Node selector' and 'Platform description' which don't apply to specific keys? I've left them as they are for now.

twaugh:
Updated.

docs/admins.rst
csomh: Yes, I saw those, I think they are find as they are.

docs/admins.rst
```diff
@@ -1,25 +1,122 @@
...
+*platform*) is specified, builds for this platform submitted to this
+cluster must include the given node selector, so as to run on a node
+of the correct architecture. This allows for installations that have
+mixed-architecture clusters and where node labels differentiate
+architecture.
+
+If the value is ``none``, this platform is the only one available and
+no node selector is required.
+
+Implementation of this requires a new optional parameter platform for
```
lcarva: This seems to describe a future behavior.

docs/admins.rst
```diff
@@ -105,6 +202,131 @@ This mechanism can also be used to temporarily disable a worker
...
+  [platform:x86_64]
+  architecture = amd64
+  enable_v1 = true
+
+  [default]
+  openshift_url = https://orchestrator.example.com:8443/
+  build_image = example.registry.com/buildroot:blue
+
+  # This node selector will be applied to triggered rebuilds:
+  low_priority_node_selector = lowpriority=true
```
lcarva: This has been removed

twaugh:
Updated.

PR #56, opened 2018-01-19
Describe lazy plugin rendering method

lcarva:
Signed-off-by: Luiz Carvalho <lucarval@redhat.com>

### DIFF
--- a/docs/build_parameters.rst
+++ b/docs/build_parameters.rst
@@ -1,0 +1,840 @@
+Build Parameters
+================
+
+OSBS’s autorebuild feature is meant to automatically start new builds of layered
+images whenever the parent image changes. This is particularly useful for image
+owners that maintain a large hierarchy of images. Currently, they need to
+manually start each image build directly and in the correct order.
+
+The goal of autorebuilds is to remove the need for image owners to start each
+required image build. Instead, they can start a build for the topmost ancestor
+which upon completion triggers the next level of layered images, and so on.
+
+In the current design of osbs-client and atomic-reactor, the parameters needed
+for rebuilding the container image are stored in the corresponding
+``BuildConfig`` object in OpenShift. It contains required information such as
+build image, atomic reactor plugins configuration, secrets to be mounted,
+``ImageChange`` triggers, git information (repository, commit, branch), and Koji
+task ID.
+
+A code release of OSBS tooling means changing the build image to be used for
+building containers.  Any new container image build will use the newer build
+image as expected. In the case of autorebuilds, the ``BuildConfig`` objects are
+not updated. When OpenShift detects a change in parent image and starts a new
+build for the layered image, it will use the outdated build image. To work
+around this limitation, image owner must start explicit builds for each image in
+the hierarchy of images to update the underlying ``BuildConfig`` object to use
+the newer build image. This contradicts our previously stated goal. This
+document outlines changes required to approach that goal.
+
+
+Design
+------
+
+There are multiple changes needed to achieve the goal of non-disruptive code
+releases. Some parts can be done by changing the process while others require
+large design changes to osbs-client and atomic-reactor.
+
+Latest Released Build Image
+"""""""""""""""""""""""""""
+
+The build image used for building container images is defined in the
+``Build``/``BuildConfig`` OpenShift objects under
+``.spec.strategy.customStrategy.from`` object. This can be a full reference to a
+specific container image in a container registry; or it can reference an
+ImageStreamTag object.
+
+To avoid having to update each ``BuildConfig`` object to use a newer build
+image, use transient tags for the build image. Transient tags are those that
+are meant to reference different images over time. To use a newer build image,
+simply move the transient tag to reference it. For instance, say we have an
+``ImageStream`` called **my-build-image** which tracks a remote container
+repository. The transient tag **released** is used to track the latest build
+image that should be used. The ``BuildConfig`` objects define the
+``ImageStreamTag`` object **my-build-image:released** is used to track the build
+image. To start using a newer build image, simply tag the newer build image with
+**released** tag.
+
+This can be done when either ``DockerImage``, or ``ImageStreamTag`` types are
+used.
+
+Environment vs User Parameters
+""""""""""""""""""""""""""""""
+
+atomic-reactor requires various parameters to build container images. These
+range from the Koji Hub URL to git commit. We will categorize them in two,
+environment and user parameters. The main difference between them is how they
+are reused. Environment parameters are shared by different build requests, while
+user parameters are unique to each build request.
+
+Environment parameters are used for configuring usage of external services such
+as Koji, Pulp, ODCS, SMTP, etc. They are also used for controlling some aspects
+of the container images built, for example, distribution scope, vendor,
+authoritative-registry, etc. These may change over time as the environment
+changes.
+
+User parameters contain the unique information for a user's build request, git
+repository, git branch, Koji target, etc. These should be reused for
+autorebuilds and are not affected by environment changes.
+
+
+Reactor Configuration
+"""""""""""""""""""""
+
+Environment configuration should be mounted in the build containers. This
+decouples the configuration from the ``Build``/``BuildConfig`` objects. The
+existing pre-build plugin ``reactor_config`` will be enhanced to provide all
+environment configuration to other plugins.
+
+Currently, the value of ``reactor_config`` is mounted into container as a
+secret. At the time it was implemented it wasn't clear how to make a
+``ConfigMap`` available in a build container. It is possible to do so by using
+an environment variable with ``valueFrom``::
+
+    apiVersion: v1
+    kind: BuildConfig
+    spec:
+      strategy:
+        customStrategy:
+          env:
+          - name: REACTOR_CONFIG
+            valueFrom:
+              configMapKeyRef:
+                name: reactor-config-map
+                key: config.yaml
+
+When the ``BuildConfig`` is instantiated, OpenShift will select the ``ConfigMap``
+of name **reactor-config-map**, retrieve the contents of the key
+**config.yaml**, and set it as the value of the **REACTOR_CONFIG** environment
+variable in the build container. For instance::
+
+    apiVersion: v1
+    kind: ConfigMap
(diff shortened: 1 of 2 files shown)


docs/build_parameters.rst
```diff
@@ -0,0 +1,833 @@
...
+        "description": "Labels that are expected to be equal in value",
+        "type": "array",
+        "items" : {
+            "type": "array",
+            "items": {
+                "type": "string"
+            }
+        }
+    },
+    "openshift": {
```
lcarva: TODO: When passed to worker builds, this must be overwritten somehow.

docs/build_parameters.rst
```diff
@@ -0,0 +1,833 @@
...
+When orchestrator build starts worker builds, it'll use the same set of secrets.
+This requires worker clusters to have the same set of secrets available. For
+example, if **reactor_config** defines::
+
+    required_secrets:
+    - kojisecret
+    - pulpsecret
+
+Secrets named kojisecret and pulpsecret must be available in orchestrator and
+worker clusters. They don't need to have the same value, just the same name.
```
twaugh: What circumstance would allow them to have different values?

docs/build_parameters.rst
```diff
@@ -0,0 +1,833 @@
...
+"""""""""""""""""""""
+
+The purpose of this new arrangement version is to easily identify whether or not
+environment parameters are provided by **reactor_config**. The order of plugins
+is not expected to change. However, hard coded, or placeholder, environment
+parameters in **orchestrator_inner** and **worker_inner** json files will
+change.
+
+A new osbs-client configuration **reactor_config_map** will be added to define
+the name of the ``ConfiMap`` object holding **reactor_config**. This
```
twaugh: s/ConfiMap/ConfigMap/

docs/build_parameters.rst
```diff
@@ -0,0 +1,833 @@
...
+configuration option will be mandatory for arrangement versions greater than or
+equal to 6. The existing osbs-client configuration **reactor_config_secret**
+will be deprecated.
```
twaugh: Warn if present when arrangement_version >= 6?

docs/build_parameters.rst
```diff
@@ -0,0 +1,833 @@
...
+
+    class OSBS(object):
+        def render_plugins_configuration(self, user_params):
+            user_params = BuildUserParams.from_json(user_params)
+
+            return PluginsConfigurationRender(user_params).render()
+
+The new class ``PluginsConfigurationRender`` will be responsible for actually
+rendering each plugin. Some of its logic will be taken from ``BuildRequest``,
+and ``DockJsonManipulator``.
```
twaugh: Duplicated, or can any be shared? (Can be sorted out during development probably)

docs/build_parameters.rst
```diff
@@ -0,0 +1,833 @@
...
+                    "krb_cache_path": {
+                        "description": "Path to kerberos credential cache file",
+                        "type": "string"
+                    },
+                    "krb_principal": {
+                        "description": "Kerberos principal",
+                        "type": "string"
+                    },
+                    "krb_keytab": {
+                        "description": "Kerberos keytab",
```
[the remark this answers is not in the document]
lcarva: The value is used as is by atomic-reactor to feed into the koji library. osbs-client's documentations says:

> location of the keytab that will be used to initialize kerberos credentials for [atomic-reactor](https://github.com/projectatomic/atomic-reactor) plugins - usually in the form `FILE:<absolute_path>`, see [kerberos documentation](http://web.mit.edu/Kerberos/krb5-latest/doc/basic/keytab_def.html) for other possible values

Maybe change description to: "Location of Kerberos keytab - e.g. FILE:<absolute_path> "
And also change the name to `krb_keytab_path` ?

docs/build_parameters.rst
```diff
@@ -0,0 +1,833 @@
...
+        }
+    },
+    "worker_token_secrets": {
+        "description": "List of OpenShift secrets for worker clusters authentication",
+        "type": "array",
+        "items": {
+            "type": "string"
+        }
+    },
+    "build_json_store": {
```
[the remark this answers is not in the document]
lcarva: osbs-client calls it `build_json_store`, but it actually reads from `build_json_dir` from configuration:

```
def get_build_json_store(self):                                                              
    return self._get_value("build_json_dir", GENERAL_CONFIGURATION_SECTION, "build_json_dir")
```

I'll change it to `build_json_dir` because that's what current admins are used to.

docs/build_parameters.rst
lcarva: We could, in theory, have different authentication certs for orchestrator and worker builds. It is unlikely that anyone would ever do that, but it's not a restriction we enforce.

docs/build_parameters.rst
lcarva: I'm wondering if we can actually use `reactor_config_map` for any arrangement, and *always* warn that `reactor_config_secret` is deprecated. What do you think?

docs/build_parameters.rst
lcarva: `DockJsonManipulater` has some logic to explicitly handle `ATOMIC_REACTOR_PLUGINS` environment variable. It might be cleaner to have its own implementation, but as you said this might be more obvious during development, yes.

docs/build_parameters.rst
```diff
@@ -0,0 +1,833 @@
...
+"""""""""""""""""""""""""""
+
+The build image used for building container images is defined in the
+``Build``/``BuildConfig`` OpenShift objects under
+``.spec.strategy.customStrategy.from`` object. This can be a full reference to a
+specific container image in a container registry; or it can reference an
+ImageStreamTag object.
+
+To avoid having to update each ``BuildConfig`` object to use a newer build
+image, use transitional tags for the build image. Transitional tags are
```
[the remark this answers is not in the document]
lcarva: Changed to transient.

docs/build_parameters.rst
```diff
@@ -0,0 +1,833 @@
...
+atomic-reactor requires various parameters to build container images. These
+range from the Koji Hub URL to git commit. We will categorize them in two,
+environment and user parameters. The main difference between them is how they
+are reused. Environment parameters are shared by different build requests, while
+user parameters are unique to each build request.
+
+Environment parameters are used for configuring usage of external services such
+as Koji, Pulp, ODCS, SMTP, etc. They are also used for controlling some aspects
+of the container images built, for example, distribution scope, vendor,
+authoritative-registry, etc. These main change over time as the environment
```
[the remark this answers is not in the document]
lcarva: Done.

docs/build_parameters.rst
```diff
@@ -0,0 +1,833 @@
...
+        "description": "Domains allowed to use when fetching artifacts by url",
+        "type": "array",
+        "items": {
+            "type": "string"
+        }
+    },
+    "image_labels": {
+        "description": "Labels to be applied to container image",
+        "type": "object",
+        "properties": {},
```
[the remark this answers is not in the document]
lcarva: I was hoping to leave this as is because the existing parameter in `add_labels_in_dockerfile` does not restrict to a known set of labels.

docs/build_parameters.rst
lcarva: Done.

docs/build_parameters.rst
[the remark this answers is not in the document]
twaugh: Sounds good, yes.

docs/build_parameters.rst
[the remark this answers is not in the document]
twaugh: Can we restrict the types though, perhaps with patternProperties?

docs/build_parameters.rst
twaugh: Yes, I like that. Simpler.

docs/build_parameters.rst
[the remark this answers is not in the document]
lcarva: Good idea, I'll look into that.

docs/build_parameters.rst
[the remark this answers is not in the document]
lcarva: Done.

docs/build_parameters.rst
lcarva: Added sentence:
> Whether functionality of ``DockJsonManipulater`` is duplicated or reused will be clearer during implementation.

docs/build_parameters.rst
lcarva: Changed text to:

> ... will be deprecated (for all arrangements).

docs/build_parameters.rst
lcarva: Added follow up sentence:

> For instance, worker and orchestrator builds may use different authentication certificates.

docs/build_parameters.rst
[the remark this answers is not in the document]
lcarva: Changed to:

```
    "image_labels": {
        "description": "Labels to be applied to container image",
        "type": "object",
        "patternProperties": {
            "^[\w\.-]+$": {"type": "string"}
        },
        "additionalProperties": false
    },
```

As per [docs](https://docs.docker.com/engine/userguide/labels-custom-metadata/):
> A label key is the left-hand side of the key-value pair. Keys are alphanumeric strings which may contain periods (.) and hyphens (-).

docs/build_parameters.rst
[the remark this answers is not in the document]
lcarva: Done.

docs/build_parameters.rst
lcarva: Added the following to the section before schema definition where `orchestrate_build` plugin is mentioned:

> The plugin will also substitute any environment configuration if needed. As an example, the **openshift** section will be replaced with the worker OpenShift cluster information.

docs/build_parameters.rst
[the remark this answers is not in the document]
twaugh: Looks good enough, although the actual rules for label names are stricter than this.

docs/build_parameters.rst
twaugh: Thanks. So, the old option name will still be accepted, but osbs-client will issue a diagnostic when it is used.

PR #59, opened 2018-04-16
Integrate ODCS design docs

lcarva:


csomh:
Reference to ODCS integration in 'Arrangement version 5' chapter should be fixed (content was moved).

docs/users.rst
```diff
@@ -113,6 +131,59 @@ not
...
+  refers to the "name" portion of the NVR (name-version-release) of an RPM, not
+  the Koji package name. Packages will be selected based on the Koji build tag
+  of the Koji build target used. The following command is useful in determining
+  which packages are available in a given Koji build tag:
+  ``koji list-tagged --inherit --latest TAG``
+
+pulp_repos
+  boolean to control whether or not an ODCS compose of type "pulp" should be
+  requested. If set to true, ``content_sets.yml`` must also be provided. A
+  compose will be requested for each architecture in ``content_sets.yml``.
```
csomh: Would be nice to have some description/example about the content of `content_sets.yml`

docs/users.rst
```diff
@@ -113,6 +131,59 @@ not
...
+
+**The "packages" and "modules" keys are mutually exclusive. If both are
+provided, "modules" will be ignored. "pulp_repos" can be used by itself, or with
+either "packages" or "modules".**
+
+
+Content Sets
+------------
+
+This files is used to define the content sets relevant to the container image.
```
csomh: Oh, here it is! :)

How about rephrasing this to say: "`content_sets.yaml is used to define...", so that when a search for `content_sets.yml` is made, this shows up as a result.

docs/users.rst
```diff
@@ -113,6 +131,59 @@ not
...
+This is relevant if RPM packages in container image are in pulp repositories.
```
csomh: This should link (the 'pulp repositories' part) to compose.pulp_repos doc.

docs/users.rst
lcarva: Fixed

docs/users.rst
lcarva: Can't link to that particular part because it's not a header. Linked to the "compose" section instead, and used text to indicate `pulp_repos` section.

lcarva:
> Reference to ODCS integration in 'Arrangement version 5' chapter should be fixed (content was moved).

Good catch! Fixed it.

docs/users.rst
csomh: A [reference](http://www.sphinx-doc.org/en/stable/markup/inline.html#role-ref) could be created—actually this is how other reference links are working—but it's ok this way :)

PR #60, opened 2018-07-31
update docs to reflect implemented designs and multistage

sosiouxme:
There was a "future designs" doc which is basically implemented at this point. I moved elements of that to other places where they made sense, or kept relevant bits in the standalone build_parameters doc. I updated the wording to reflect usage rather than design.

And of course, this enabled discussing imagebuilder and multistage.

I would love to also fill in more of the admin and user sections but... not this PR.


docs/admins.rst
```diff
@@ -113,6 +119,53 @@ the x86_64 platform will publish v1 images as well as v2 images. Other
...
+object to use the newer build image. But this is a tedious and error-prone
+approach.
+
+Latest Released Build Image
+'''''''''''''''''''''''''''
+
+To avoid having to update each ``BuildConfig`` object to use a newer build
+image, the build image should be specified with a transient tag. Transient
+tags are those that are meant to reference different images over time. To use a
+newer build image, simply move the transient tag to reference it. For instance,
```
sosiouxme: @lcarva @rcerven is this accurate? Do we actually do this? I wasn't sure if OpenShift would check that the tag has been updated and now points at a new buildroot.

docs/users.rst
```diff
@@ -170,9 +170,38 @@ be provided, or "pulp_repos" set to true. Otherwise, build will fail.**
...
 
+.. _container.yaml-autorebuild:
+
+autorebuild
+~~~~~~~~~~~
+
+This section specifies whether and how a build should be rebuilt based on
+changes to the base parent image.
+
+TODO
```
twaugh: Is this a place-holder for filling out more about container.yaml keys?

docs/admins.rst
lcarva: I don't think OpenShift will do that. There's a way to set the policy so it does check for updates, but I think that might be available just for pods, not builds.

docs/admins.rst
```diff
@@ -345,3 +345,108 @@ containing "enable_v1 = true".
...
+Priority of Container Image Builds
+----------------------------------
+
+For a build system it's desirable to prioritize different kinds of builds in
+order to better utilize resources. Unfortunately, OpenShift's scheduling
+algorithm does not support setting a priority value for a given build. To
+achieve some sort of build prioritization, we can leverage node selectors to
+allocate different resources to different build types.
+
+Consider the following types of container builds builds:
```
lcarva: "builds builds"

docs/admins.rst
```diff
@@ -345,3 +345,108 @@ containing "enable_v1 = true".
...
+
+The task load of the Koji builders used by OSBS will not reflect the actual load
+on the OpenShift cluster used by OSBS. The disparity is due to auto rebuilds not
+having a corresponding Koji task. This creates a scenario where a buildContainer
+Koji task is started, but the OpenShift build remains in pending state. The Koji
+builder capacity should be set based on how many nodes allow **scratch builds**
+and/or **explicit builds**. In the example above, there are 4 nodes that allow
+such builds.
+
+Providing the osbs-client logs in the Koji task should give users a better
```
lcarva: The log file, `osbs-client.log`, in a Koji task give users a better understanding of any delays due to scheduling.

docs/admins.rst
```diff
@@ -345,3 +345,108 @@ containing "enable_v1 = true".
...
+instance. In other cases, it could be due to an in-progress feature. For this
+reason, it cannot be said that all *explicit builds* are higher priority than
+*auto rebuilds*, or vice-versa.
+
+However, *auto rebuilds* have the potential of completely consuming OSBS’s
+infrastructure. There must be some mechanism to throttle the amount of *auto
+rebuilds*. For this reason, OSBS uses a different node selector for each
+different build type:
+
+- *scratch build*: builds_scratch=true
```
lcarva: This is actually configured in osbs-client secret. It can be named whatever the user wants. It's beyond the scope of this change though.

docs/admins.rst
```diff
@@ -24,11 +30,11 @@ The parameter ``can_orchestrate`` defaults to false. The API method
...
 
-reactor_config_secret
+reactor_config_map
 ~~~~~~~~~~~~~~~~~~~~~
 
-When ``reactor_config_secret`` is specified this is the name of a
-Kubernetes secret holding :ref:`config.yaml`. A pre-build plugin will
+``reactor_config_map`` specifies the name of a
+Kubernetes configmap holding :ref:`config.yaml`. A pre-build plugin will
 be configured with the location this secret is mounted.
```
lcarva: A pre-build plugin will read its value from `REACTOR_CONFIG` environment variable.

docs/admins.rst
```diff
@@ -113,6 +119,53 @@ the x86_64 platform will publish v1 images as well as v2 images. Other
 platforms on instance1, and all platforms on instance2, will only
 publish v2 images.
 
+Autorebuilds and OSBS updates
```
lcarva: I think most of the text in this section can be removed. It was helpful in helping providing background to implement support for reactor config map.

:bomb: it?

The bit about using ImageStreams instead of images directly is still relevant.

docs/users.rst
```diff
@@ -182,6 +182,26 @@ TODO
 
 .. _content_sets.yml:
 
+image_build_method
```
lcarva: Should we add this to the example above (L92-115)?

.gitignore
```diff
@@ -1,2 +1,3 @@
 _build*
 _build_html*
+**/*.swp
```
lcarva: Since I've never seen vim crash, I've had this in my`.vimrc` for years: `set noswapfile` :smile_cat:

sosiouxme: vim may not crash, but my laptop sure does...

docs/admins.rst
sosiouxme: Eh, alright. But...

docs/admins.rst
sosiouxme: Are we currently just letting autorebuilds use old buildroots?

The transient tag approach *would* work, if just combined with removing the previous tagged version from docker after a code update...

But for us specifically I imagine we want to record exactly which buildroot was used.

docs/admins.rst
sosiouxme: Are these at least the defaults though?

docs/users.rst
sosiouxme: Well, more about autorebuilds specifically (got any seed info?), but it would be nice to describe all available members.

docs/users.rst
sosiouxme: Reckon so :) :+1:

docs/users.rst
twaugh: Unfortunately the [schema|https://github.com/projectatomic/atomic-reactor/blob/1.6.33/atomic_reactor/schemas/container.json#L21] is silent on the matter, but the [code|https://github.com/projectatomic/atomic-reactor/blob/1.6.33/atomic_reactor/plugins/pre_check_and_set_rebuild.py#L112] says there is a single key, "from_latest". @lcarva, is that described somewhere?

lcarva: No, I don't think it's externally described anywhere.

docs/admins.rst
twaugh: Actually there are no defaults. If no node selector is specified for a given type, no node selector is used when submitting builds of that type.

docs/users.rst
twaugh: Mentioned here: https://github.com/projectatomic/atomic-reactor/pull/1056

docs/build_process.rst
```diff
@@ -359,3 +378,18 @@ Note:
...
+
+OSBS’s autorebuild feature automatically starts new builds of layered images
+whenever the base parent image changes. This is particularly useful for image
+owners that maintain a large hierarchy of images, which would otherwise require
+manually starting each image build in the correct order.  Instead, image owners
+can start a build for the topmost ancestor which upon completion triggers the
+next level of layered images, and so on.
+
+Builds may opt in to autorebuilds with an
+:ref:`autorebuild entry in the dist-git configuration. <container.yaml-autorebuild>`
```
lkolacek: @twaugh just a quick question should we update our chainrebuilds test to use container.yaml instead of .osbs-repo-config ??

lkolacek:
Documentation looks great, thanks Luke!

docs/users.rst
```diff
@@ -170,9 +172,38 @@ be provided, or "pulp_repos" set to true. Otherwise, build will fail.**
 provided, "modules" will be ignored. "pulp_repos" can be used by itself, or with
 either "packages" or "modules".**
 
+.. _container.yaml-autorebuild:
+
+autorebuild
+~~~~~~~~~~~
+
+This section specifies whether and how a build should be rebuilt based on
```
twaugh: I don't think it specifies "whether" -- that comes from a file named ".osbs-repo-config" in the dist-git tree:
```
[autorebuild]
enabled = true|false  # defaults to false
```
ref: https://github.com/projectatomic/osbs-client/blob/0.49/osbs/repo_utils.py#L63

docs/build_process.rst
twaugh: No. The "whether" configuration comes from .osbs-repo-config, but the "how" comes from container.yaml.

lkolacek: I see thanks

PR #61, opened 2018-08-22
Update PR review checklist (OSBS-5770)

twaugh:

### DIFF
--- a/docs/contributors.rst
+++ b/docs/contributors.rst
@@ -34,11 +34,20 @@ the pull request is merged.
 
 Before a pull request is approved it must meet these criteria:
 
-- unit tests pass
+- Commit messages are descriptive enough
 
-- code coverage from testing does not decrease and new code is covered
+- "Signed-off-by:" line is present in each commit
 
-- new features must have user documentation
+- Unit tests pass
+
+- Code coverage from testing does not decrease and new code is covered
+
+- JSON/YAML configuration changes are updated in the relevant schema
+
+- Changes to metadata also update the documentation for the metadata
+
+- Pull request includes link to an osbs-docs PR for user documentation
+  updates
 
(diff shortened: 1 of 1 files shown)


docs/contributors.rst
```diff
@@ -34,11 +34,22 @@ the pull request is merged.
...
+- "Signed-off-by:" line is present in each commit
 
-- new features must have user documentation
+- Copyright notice for each modified file includes current year
+
+- Unit tests pass
+
+- Code coverage from testing does not decrease and new code is covered
+
+- JSON configuration changes are updated in the relevant schema
```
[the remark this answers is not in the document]
twaugh: Thanks, fixed.

docs/contributors.rst
```diff
@@ -34,11 +34,20 @@ the pull request is merged.
...
 
-- code coverage from testing does not decrease and new code is covered
+- "Signed-off-by:" line is present in each commit
 
-- new features must have user documentation
+- Unit tests pass
+
+- Code coverage from testing does not decrease and new code is covered
+
+- JSON/YAML configuration changes are updated in the relevant schema
```
mlangsdorf: Is there a way to detect schema changes and catch that they haven't been updated, in the same way that the linter catches format errors and Travis runs the unit tests and helps calculate code coverage?

twaugh: Not really. A configuration change could be something like a new call to `data.get('use_latest')`, and it would be very hard to work out whether that's coming from something a schema is supposed to validate, or whether it is represented in the appropriate schema.

Right now we have unqualified "object" types in JSON Schema definitions, i.e. saying it's an object without saying anything further about what properties it has, so technically the schema definitions can validate but they don't do it as usefully as they could.

PR #64, opened 2018-11-27
empty packages behavior

mlangsdorf:
Describe the new behavior with an empty packages section.

### DIFF
--- a/docs/users.rst
+++ b/docs/users.rst
@@ -148,7 +148,8 @@ packages
   of the Koji build target used. The following command is useful in determining
   which packages are available in a given Koji build tag:
   ``koji list-tagged --inherit --latest TAG``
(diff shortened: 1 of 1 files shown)


docs/users.rst
```diff
@@ -148,7 +148,8 @@ packages
   of the Koji build target used. The following command is useful in determining
   which packages are available in a given Koji build tag:
   ``koji list-tagged --inherit --latest TAG``
-
+  If packages section is declared but is empty, the compose will include all
```
twaugh: Double-quotes around `packages` for consistency with the other added paragraph?

mlangsdorf: Fixed, and declaration changed to key for similar consistency.

PR #65, opened 2018-11-29
add some project links

ktdreyer:
Link GitHub users to the main page where readthedocs renders OSBS's documentation so this is easier to find. 

Link the main index page to the koji-containerbuild project so it's easier to find that as well.

~While we're here, koji-containerbuild contains multiple plugins, so pluralize that description.~

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -4,7 +4,8 @@ OpenShift Build Service
 
 OSBS is a collection of tools, workflows and integration points that build and release layered container images.
 
-OSBS hooks into Koji_ with the help of the koji-containerbuild plugin, and
+OSBS hooks into Koji_ with the help of the `koji-containerbuild plugin
+<https://github.com/release-engineering/koji-containerbuild>`_, and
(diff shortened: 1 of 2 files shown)


docs/index.rst
```diff
@@ -4,7 +4,8 @@ OpenShift Build Service
 
 OSBS is a collection of tools, workflows and integration points that build and release layered container images.
 
-OSBS hooks into Koji_ with the help of the koji-containerbuild plugin, and
+OSBS hooks into Koji_ with the help of the `koji-containerbuild plugins
```
athos-ribeiro: If we are to refer to koji-container build as a set of plugins, perhaps we should do it everywhere. Namely, the koji-containerbuild repository as well.

twaugh: There are two plugins in the koji-containerbuild project: one for the CLI tool, and one for the builder. For the purpose of this sentence, OSBS uses the help of one of those plugins (the builder one). So I think singular makes sense here.

ktdreyer: To ease the acceptance of this particular pull request, I've deleted the final commit that pluralizes "plugins" here :) It's really a separate issue.

PR #71, opened 2019-03-12
inherit yum repositories and odcs composes from baseimage build

rcerven:
* OSBS-6662

### DIFF
--- a/docs/users.rst
+++ b/docs/users.rst
@@ -113,6 +113,8 @@ An example::
     - "module_name2:stream1"
     # Possible values, and default, are configured in OSBS environment.
     signing_intent: release
+    # used for inheritance of yum repos and ODCS composes from baseimage build
+    inherit: true
 
    image_build_method: docker_api
 
@@ -173,14 +175,16 @@ signing_intent
   environment. See :ref:`config.yaml-odcs` section for environment configuration
   details, and full explanation of :ref:`signing-intent`.
 
-**If "compose" section is defined, then "pulp_repos" must be set to true or there
-must be a valid "modules" or "packages" key.  If there is a "modules" key, it
-must have a non-empty list of modules.  The "packages" key, and only the "packages"
+inherit
+  boolean to control whether or not to inherit yum repositories and odcs composes
+  from baseimage build, default false. Scratch and isolated builds do not support
+  inheritance and false is always assumed.
+
+**If there is a "modules" key, it
+must have a non-empty list of modules. The "packages" key, and only the "packages"
 key, can have an empty list.**
 
-**The "packages" and "modules" keys are mutually exclusive. If both are
-provided, "modules" will be ignored. "pulp_repos" can be used by itself, or with
-either "packages" or "modules".**
+**The "packages", "modules" and "pulp_repos" keys can be used mutually.**
 
 .. _container.yaml-autorebuild:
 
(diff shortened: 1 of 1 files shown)


docs/users.rst
```diff
@@ -173,14 +175,16 @@ signing_intent
   environment. See :ref:`config.yaml-odcs` section for environment configuration
   details, and full explanation of :ref:`signing-intent`.
 
-**If "compose" section is defined, then "pulp_repos" must be set to true or there
-must be a valid "modules" or "packages" key.  If there is a "modules" key, it
-must have a non-empty list of modules.  The "packages" key, and only the "packages"
+inherit
+  boolean to control whether or not to inherit yum repositories and odcs composes
+  from baseimage build, default false. Inheritance works only for builds
+  which aren't scratch or isolated.
```
athos-ribeiro: What is the behavior here when it is set to `true` in a scratch/isolated build? If it is just ignored, maybe we could say something like

> ... from base image build, defaults to false. Scratch and isolated builds do not support iInheritance and `false` is always assumed.

docs/users.rst
```diff
@@ -173,14 +175,16 @@ signing_intent
...
+
+**If there is a "modules" key, it
+must have a non-empty list of modules. The "packages" key, and only the "packages"
 key, can have an empty list.**
 
-**The "packages" and "modules" keys are mutually exclusive. If both are
-provided, "modules" will be ignored. "pulp_repos" can be used by itself, or with
-either "packages" or "modules".**
+**The "packages" and "modules" and "pulp_repos" keys can be used mutually.**
```
athos-ribeiro: **The "packages", "modules", and "pulp_repos" keys can be used mutually.**

docs/users.rst
```diff
@@ -493,7 +497,7 @@ Yum repository URL
 
 As part of a build request, you may provide the ``repo-url`` parameter with the
 URL to a yum repository file. This file is injected into the container build.
-Note that "ODCS compose" method is disabled if this parameter is given.
+Note that it will be used together with "ODCS compose" method.
```
athos-ribeiro: Can we rephrase this or maybe ommit the sentence?

docs/users.rst
```diff
@@ -503,6 +507,13 @@ yum repository for the build tag of target is automatically injected in
 container build. This behavior is disabled if either "ODCS compose" or "Yum
 repository URL" are used.
 
+Inherited yum repository and ODCS compose
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If you want to inherit yum repositories and ODCS composes from baseimage build,
+you can enable it via the "inherit" key under "compose" in ``container.yaml``.
+Works for builds which aren't scratch or isolated.
```
athos-ribeiro: What about

> Does not support scratch or isolated builds.

?

docs/users.rst
rcerven: I want to mention it, that it can now be used together, rephrased a bit

athos-ribeiro: Oh, I see. What would you think about something like

> Since aromic-reactor x.y.z, it is possible to combine ODCS composes with  repository files. In previous versions, the ODCS compose would be disabled if a repository file URL was given.

Then it would be clear that this was not supported before, and it is supported now. If you don't want to mention atomic-reactor in the user docs, it could also go in the lines of 

> Current OSBS versions support the combination of ODCS composes with  repository files. This is a change to OSBS former behavior, where the ODCS compose would be disabled if a repository file URL was given.

PR #73, opened 2019-04-11
Mention test.sh in contributors.rst

twaugh:


twaugh:
When we merge this, let's check to see whether the change is reflected here: https://osbs.readthedocs.io/en/latest/contributors.html

I removed the old broken integration and set up a new one using a webhook. Hopefully it will update automatically again now.

PR #75, opened 2019-05-16
Koji/registry digests verification

athos-ribeiro:


docs/users.rst
```diff
@@ -440,6 +440,23 @@ label which causes OSBS to query koji for what the next release value should be.
 This is done via koji's ``getNextRelease`` API method. In either case, the
 release value can also be overridden by using the ``release`` API parameter.
 
+During the build process, OSBS will query koji for the builds of all parent
+images using their NVRs. If any of the parent image builds is not found in
+koji, or if NVR information cannot be extracted from the parent image, OSBS
+assumes that the parent image was not built by OSBS and halts the current
+build. In othe words, an image cannot be built using a parent image which has
+not been built by OSBS.
```
MartinBasti: Maybe add note somehwere that this feature can be disabled?

athos-ribeiro: I have been thinking about it. But then, this note should go in the administration page somewhere (note that we already have his in the config schema which is linked there).
Maybe we should duplicate part of this explanation in the administration page and describe the option to disable the feature. Thoughts?

MartinBasti: I'd not duplicate that or complicated that. I just want to make users aware that this functionality may be disabled in config and admin may look for that option in config

athos-ribeiro: Added a comment on the possibility of disabling it though reactor config and added a link to the config.json schema

PR #76, opened 2019-05-28
Remove references to v1 content pushes to Pulp

athos-ribeiro:
* OSBS-6665


rcerven:
more to delete/change from example bellow that enable_v1 (you've deleted)

PR #78, opened 2019-06-03
Update the configmap example with Koji's fast upload option

cverna:
Signed-off-by: Clement Verna <cverna@tutanota.com>

### DIFF
--- a/docs/build_parameters.rst
+++ b/docs/build_parameters.rst
@@ -83,6 +83,7 @@ Example of **REACTOR_CONFIG**::
         root_url: https://koji.example.com/root
         auth:
             ssl_certs_dir: /var/run/secrets/atomic-reactor/kojisecret
+        use_fast_upload: false
 
     pulp:
         name: my-pulp


twaugh:
That doesn't seem valid according to the [schema](https://github.com/containerbuildsystem/atomic-reactor/blob/master/atomic_reactor/schemas/config.json).

cverna:
Coresponding PR --> https://github.com/containerbuildsystem/atomic-reactor/pull/1220

docs/build_parameters.rst
```diff
@@ -83,6 +83,7 @@ Example of **REACTOR_CONFIG**::
         root_url: https://koji.example.com/root
         auth:
             ssl_certs_dir: /var/run/secrets/atomic-reactor/kojisecret
+        use_fast_upload: true
```
MartinBasti: should we recommend `True` here? Is that only current fedora-infra workaround or a preferred way of uploading?

cverna: I am honestly not sure, but `use_fast_upload` defaults to true in the koji.conf you get when installing the koji client. 
Generally I think setting it to true will not hurt :smile:, but I am happy to change that to `false` if needed.

twaugh: I'd actually be happy for us to default to fast-upload, i.e. for this option to have default value `true`.

MartinBasti: ACK

twaugh:
Now the default is true, the only useful value in the config map is false. Should we change that in this example?

cverna:
> Now the default is true, the only useful value in the config map is false. Should we change that in this example?

Updated :+1:

PR #81, opened 2019-06-18
Describe how floating tags are chosen

twaugh:

### DIFF
--- a/docs/users.rst
+++ b/docs/users.rst
@@ -400,11 +400,18 @@ Dockerfile, the tags used within the repository are:
 
 - a unique tag including the timestamp (this tag is the only tag
   applied for scratch builds)
-- ``latest``
+- ``{version}-{release}`` (the ``version`` and ``release`` labels
+  together)
+
+If the ``tags`` value is set in container.yaml, those tags are applied
+to the image as floating tags.
+
+If ``tags`` is not used in container.yaml, the following tags are
+applied:
+
 - ``{version}`` (the ``version`` label)
-- ``{version}-{release}`` (the ``version`` and ``release`` labels together)
-- any additional tags configured in the git repository (named in the
-  ``additional-tags`` file)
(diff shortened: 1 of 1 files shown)


docs/users.rst
```diff
@@ -398,11 +398,17 @@ Dockerfile, the tags used within the repository are:
...
+If the ``tags`` value is set in container.yaml, those tags are applied
+to the image as floating tags.
+
+If ``tags`` is not used in container.yaml, the following tags are
+applied:
 - ``{version}`` (the ``version`` label)
-- ``{version}-{release}`` (the ``version`` and ``release`` labels together)
-- any additional tags configured in the git repository (named in the
-  ``additional-tags`` file)
+- ``latest``
```
MartinBasti: This is not showing a list with bullets but inline on single line in GH view. (GH issue?)
IMHO it needs extra newline after `applied:`

twaugh: Thanks, should be fixed now.

docs/users.rst
```diff
@@ -398,11 +398,18 @@ Dockerfile, the tags used within the repository are:
...
+
+If ``tags`` is not used in container.yaml, the following tags are
+applied:
+
 - ``{version}`` (the ``version`` label)
-- ``{version}-{release}`` (the ``version`` and ``release`` labels together)
-- any additional tags configured in the git repository (named in the
-  ``additional-tags`` file)
+- ``latest``
+- any additional tags named in the ``additional-tags`` file
```
ktdreyer: Tim, I was wondering about this `additional-tags` file. Is this a legacy thing that users should not use anymore?

twaugh: It's certainly less flexible. We haven't deprecated it yet though.

twaugh: Deprecation tracked in OSBS-7698.

PR #85, opened 2019-07-02
OMPS integration

MartinBasti:
* OSBS-7542

Signed-off-by: Martin Bašti <mbasti@redhat.com>

### DIFF
--- a/docs/users.rst
+++ b/docs/users.rst
@@ -759,3 +759,13 @@ releases.
 
 
 .. _`config.json`: https://github.com/containerbuildsystem/atomic-reactor/blob/master/atomic_reactor/schemas/config.json
+
+After a successful build, if :ref:`OMPS integration <omps-integration>`
+is enabled, operator manifests are uploaded into configured application
+registry and namespace.
+
+Details on how operator manifest can be accessed from the application registry are
+stored in koji build, in section ``build.extra.operator_manifests.appregistry``.
+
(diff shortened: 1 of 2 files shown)


docs/users.rst
```diff
@@ -759,3 +759,13 @@ releases.
 
 
 .. _`config.json`: https://github.com/containerbuildsystem/atomic-reactor/blob/master/atomic_reactor/schemas/config.json
+
+With enabled  :ref:`OMPS integration <omps-integration>` operator manifests are
```
MartinBasti: This ref is not processed correctly in github view, I'm not sure how to fix it

athos-ribeiro: It does render properly when building html docs though. Should not be a problem at all.

docs/users.rst
```diff
@@ -759,3 +759,13 @@ releases.
...
+after successful build uploaded into configured application registry and
+namespace.
+
+Details how operator manifest can be accessed from application registry are
+stored in koji build, in section ``build.extra.operator_manifests.appregistry``.
+
+Manifests from scratch, isolated and re-builds are not pushed into application
```
athos-ribeiro: How about: Manifests will not be pushed to the application registry for scratch builds, isolated builds, or re-builds to prevent unwanted changes ?

MartinBasti:
@athos-ribeiro thank you! updated

PR #99, opened 2019-10-17
users: describe "unique", "primary", and "floating" tags

ktdreyer:
Rewrite the "Image tags" section to describe the three categories of tags that OSBS uses: "unique", "primary", and "floating".

Drop the first paragraph from the "Isolated Builds" section since it duplicates the Image tags section above. Use the term "floating tags" here instead. This should hint to the user to scroll up and read about what a "floating tag" is.

Update the final paragraph in the "Isolated Builds" section so that our definitions of "primary tag" and "unique tag" match what the "Image tags" section.

### DIFF
--- a/docs/users.rst
+++ b/docs/users.rst
@@ -729,28 +729,37 @@ include the fetched artifacts in addition to the installed RPMs.
 Image tags
 ----------
 
-The output from atomic-reactor includes container images tagged into a
-registry (or Pulp, if Pulp integration is enabled). In addition, when
-multi-platform builds are enabled each set of images will be grouped
-into a manifest list, which itself is tagged.
+OSBS's atomic-reactor pushes the new container image to the container
+registry (or Pulp, if Pulp integration is enabled) and updates various tag
+references in the registry. In addition, when multi-platform builds are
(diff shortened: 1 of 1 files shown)


ktdreyer:
I'm gonna need to rebase this after https://github.com/containerbuildsystem/osbs-docs/pull/96 lands.

docs/users.rst
```diff
@@ -728,28 +728,37 @@ include the fetched artifacts in addition to the installed RPMs.
...
-Dockerfile, the tags used within the repository are:
+OSBS determines the name of the repository from the ``name`` label in the
+Dockerfile. There are three categories of tags that OSBS creates when
+tagging the resulting image in the registry:
 
-- a unique tag including the timestamp (this tag is the only tag
-  applied for scratch builds)
-- ``{version}-{release}`` (the ``version`` and ``release`` labels
-  together)
+* A "unique" tag: This tag includes the timestamp of when the image was
```
MartinBasti: Maybe we could make tag type more visible by italics+bold font?

ktdreyer: Good idea. I've made the tags bold and amended this pull request.

PR #103, opened 2019-10-31
Add initial travis CI setup

athos-ribeiro:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: python
+
+branches:
+  only:
+    - master
+
+install:
+  - pip install -r docs/requirements.txt
+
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,13 @@
+language: python
+
+branches:
+  only:
+    - master
```
MartinBasti: why not all?

twaugh: It's to avoid 'double builds on pull requests': https://docs.travis-ci.com/user/pull-requests#double-builds-on-pull-requests

branches.only controls builds in response to merges. Builds in response to pull requests are set in Travis CI settings.

PR #104, opened 2019-11-01
add OSBS structure diagram to index page

ktdreyer:
This picture shows how the OSBS components fit together.


athos-ribeiro:
This looks great. Thank you! Could we refer to the `atomic-reactor container` as `buildroot container` which has both osbs-client and atomic-reactor installed (instead of just osbs-client)?

ktdreyer:
Sure, I've amended this with the buildroot container.

PR #115, opened 2019-11-19
users: align content_sets.yml anchor with "Content Sets" section

ktdreyer:
Prior to this change, the content_sets anchor was pointing at the image_build_method section. This caused Sphinx to generate links like "See image_build_method" instead of "See content_sets.yml".


ktdreyer:
I noticed this because  the ["pulp_repos" description](https://osbs.readthedocs.io/en/latest/users.html#compose) says "See image_build_method."

PR #117, opened 2019-11-25
set_release_env: add documentation for set_release_env

mlangsdorf:

### DIFF
--- a/docs/users.rst
+++ b/docs/users.rst
@@ -490,6 +490,17 @@ version
 This key is no longer used by OSBS and is only kept in the schema for backwards
 compatibility.
 
+
+set_release_env
+~~~~~~~~~~~~~~~
+
+Optional string.  If set, osbs-client will modify each stage of the image's 
+Dockerfile, adding an ENV statement immediately following the FROM statement.
+The ENV statement will assign an environment variable with the same name as
+the value of set_release_env and the value of the current build's release number.
+Users can use this environment variable to get the release value when running
+tools inside the container.
+
 .. _container.yaml-autorebuild:
 
 autorebuild


docs/users.rst
```diff
@@ -490,6 +490,16 @@ version
 This key is no longer used by OSBS and is only kept in the schema for backwards
 compatibility.
 
+
+set_release_env
+~~~~~~~~~~~~~~~
+
+Optional string.  If set, the image's Dockerfile will be amended with
```
twaugh: "amended with": can we say it will be added to the start of each stage in the Dockerfile, directly after FROM? (can be clarified in a separate paragraph if that makes it easier)

I think it's useful to be explicit about which stages we'll modify here.

docs/users.rst
```diff
@@ -490,6 +490,16 @@ version
...
+"ENV $set_release_env=$RELEASE" and an environment variable with of the string
```
twaugh: "ENV $set_release_env=$RELEASE": it might read as though we add that literal string, which of course we don't. Can we say "an ENV statement"? The words afterwards clarify what the statement does.

"with of": some words missing?

docs/users.rst
mlangsdorf: Clarified that we're editing all of them.

docs/users.rst
mlangsdorf: Rewritten, thanks for the review.

docs/users.rst
```diff
@@ -490,6 +490,17 @@ version
 This key is no longer used by OSBS and is only kept in the schema for backwards
 compatibility.
 
+
+set_release_env
+~~~~~~~~~~~~~~~
+
+Optional string.  If set, osbs-client will modify each stage of the image's 
+Dockerfile, adding an ENV statement at the end of the FROM section.  The
```
twaugh: Maybe instead of "at the end of the FROM section" it should be "immediately following the FROM statement". Then image owners know they are free to use that environment variable at any time.

PR #121, opened 2019-12-11
remove references to pdc support

mlangsdorf:
pdc support was deprecated in September 2018, so remove the last remnants
of it from the documentation.


MartinBasti:
Nitpick: maybe `pdc` could be removed from `docs/.aspell.en.pws` too

PR #124, opened 2020-01-07
Describe usage of remote source build args

athos-ribeiro:
Describe how the build args provided by the remote source plugins may be
used to extract and copy the remote sources into the build container.

* OSBS-8473

### DIFF
--- a/docs/users.rst
+++ b/docs/users.rst
@@ -601,7 +601,7 @@ Fetching source code from external source using cachito
 
 As described in :ref:`cachito-integration`, it is possible to use cachito to
 download a tarball with an upstream project and its dependencies and make it
-available for usage in the working directory used during an OSBS build.
+available for usage during an OSBS build.
 
 remote_source
 ~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


docs/users.rst
```diff
@@ -629,6 +629,27 @@ flags
...
   for further reference.
 
+Once the `remote_source` map described above is set in ``container.yaml``, you
+can now extract the upstream sources (with bundled dependencies) archive
+provided by cachito in your build image by adding::
+
+    RUN mkdir -p $REMOTE_SOURCE_DIR
+    ADD $REMOTE_SOURCE $REMOTE_SOURCE_DIR
+
+to your Dockerfile.
```
MartinBasti: I'm curious if OSBS could do this Dockerfile change for user

athos-ribeiro: The tricky part is that this is a multistage build. If we do so, should we add them in the first stage? Should we remove the sources dir in the final stage as well?

MartinBasti: I see.
In multistage builds we don't have to remove anything, builds will be discarded unless user copy results to final build

PR #128, opened 2020-02-03
make source containers aware of cachito

rcerven:
* OSBS-8233

### DIFF
--- a/docs/users.rst
+++ b/docs/users.rst
@@ -1248,8 +1248,10 @@ available in the OSBS buildroot as `bsi` executable in `$PATH`.
 
 
 Current limitations:
-* only Source RPMs are added into source container image
+
+* only Source RPMs and sources fetched through :ref:`cachito-integration` are added into source container image
 * only koji internal RPMs are supported
+
 Support for other types of sources and external builds will be added in future.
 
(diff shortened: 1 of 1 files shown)


docs/users.rst
```diff
@@ -1248,8 +1248,10 @@ available in the OSBS buildroot as `bsi` executable in `$PATH`.
 
 
 Current limitations:
-* only Source RPMs are added into source container image
+
+* only Source RPMs and sources fetched through cachito integration are added into source container image
```
MartinBasti: Is possible to link `cachito` to cachito section?

rcerven: added

docs/users.rst
```diff
@@ -1248,8 +1248,10 @@ available in the OSBS buildroot as `bsi` executable in `$PATH`.
 
 
 Current limitations:
-* only Source RPMs are added into source container image
+
+* only Source RPMs and sources fetched through :ref:`cachito-integration` are added into source container image
```
athos-ribeiro: We should substitute the "*" for "-" in both lines here: This is how lists in rst work.

athos-ribeiro: NVM, maybe it was broken due to the missing line break!

PR #131, opened 2020-02-13
Document cachito dependency replacement option

athos-ribeiro:
* OSBS-8139

### DIFF
--- a/docs/users.rst
+++ b/docs/users.rst
@@ -656,6 +656,46 @@ after the build is complete::
 ``$REMOTE_SOURCE`` is another build arg, which points to the extracted tar
 archive provided by cachito in the buildroot workdir.
 
+Replacing project dependencies with cachito
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Cachito also provides a feature to allow users to replace a project's
+dependencies with another version of that same dependency or with a completely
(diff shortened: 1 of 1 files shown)


docs/users.rst
```diff
@@ -656,6 +656,46 @@ after the build is complete::
...
+  koji container-build [...] --scratch --replace-dependency gomod:pagure.org/cool-go-project:v1.2 --replace-dependency gomod:gopkg.in/foo:2:github.com/bar/foo
+
+In the examples above, two dependencies would be replaced. cool-go-project
+would be used in version ``v1.2``, no matter what version is specified by the
+project requesting it. Whereas ``gopkg.in/foo`` will be replaced by
+``github.com/bar/foo`` version 2.
+
+Note that while in ``fedpkg`` the replace dependency option receives multiple
+parameters, the same option should be specified multiple times in ``koji`` or
+the ``osbs`` CLI. This was done to keep the consistency with the similar option
```
rcerven: not sure if we have to mention : This was done to keep the consistency ....

athos-ribeiro: I was asking myself if I shouldn't have gone for multiple --opts in fedpkg as well... I'd rather just be clear to the users here (instead of have to say that whenever someone points it out). It may also be useful to the future me in case I forget why we did it this way... ^^'

PR #135, opened 2020-04-03
Prevent building source containers if lookaside is involved

rcerven:
* OSBS-8774


docs/users.rst
```diff
@@ -1295,6 +1295,7 @@ previously created by OSBS. To create a source container build you have to
 specify either koji N-V-R or build ID for the image build you want to
 create a source container image for.
 
+Source container build will fail if koji build was using lookaside cache.
```
athos-ribeiro, line 1298: Could we add the reason behind it?

PR #136, opened 2020-04-07
Fix links to OKD docs

MartinBasti:
Documentation pinned to okd 3.11 as latest version changes

Signed-off-by: Martin Bašti <mbasti@redhat.com>


MartinBasti:
OMG

Issue #118, opened 2019-11-26
"Override Parent Image" does not describe multi-stage builds


ktdreyer:
This section describes how we can dynamically override the parent image:

https://osbs.readthedocs.io/en/latest/users.html#override-parent-image

It does not mention how this interacts with multi-stage builds. In particular, I want to update the *final* image's `FROM` line, but not the earlier builder `FROM` lines, and the docs clearly tell me OSBS's behavior here.

ktdreyer:
(cc'ing @b-ranto since we were looking at this recently)

rcerven:
it updates just the final parent image

PR #139, opened 2020-04-15
users: describe koji_parent_build for multi-stage builds

ktdreyer:
The first commit rewrites the first paragraph in the "Override Parent Image" section to use the active voice so that users can understand it more easily.

The second commit clarifies that OSBS will only override the final "FROM" instruction for a multi-stage build.

Fixes: #118

### DIFF
--- a/docs/users.rst
+++ b/docs/users.rst
@@ -895,12 +895,12 @@ build.
 Override Parent Image
 ----------------------
 
-The parent image used for building a layered image is determined by the ``FROM``
-instruction in the Dockerfile by default. Users can override this behavior
-by specifying a koji parent build via the ``koji_parent_build`` API parameter.
-When given, the image reference in the provided koji parent build will be used as
-the value of the FROM instruction. The same source registry restrictions
-apply.
+OSBS uses the ``FROM`` instruction in the Dockerfile to find a parent image
+for a layered image build. Users can override this behavior by specifying a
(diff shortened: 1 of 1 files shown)


docs/users.rst
```diff
@@ -895,12 +895,14 @@ build.
...
-The parent image used for building a layered image is determined by the ``FROM``
-instruction in the Dockerfile by default. Users can override this behavior
-by specifying a koji parent build via the ``koji_parent_build`` API parameter.
-When given, the image reference in the provided koji parent build will be used as
-the value of the FROM instruction. The same source registry restrictions
-apply.
+OSBS uses the ``FROM`` instruction in the Dockerfile to find a parent image
+for a layered image build. Users can override this behavior by specifying a
+koji parent build via the ``koji_parent_build`` API parameter in the
+``buildContainer`` task. When a user specifies a ``koji_parent_build``
```
athos-ribeiro, line 901: The `koji_parent_build` can also be passed through the osbs cli or api (wihtout going through a koji  task or the koji-containerbuild plugin). This is not the most straightforward way of triggering a build, but could still be used. Since this is the upstream documentation,could we rephrase this bit to make that clear?

ktdreyer: Oh, yeah, I should not have added the `buildContainer task` phrase. I've deleted it and re-pushed this branch.

PR #140, opened 2020-04-16
Lint all Markdown files and enable CI MD lint

ben-alkov:


MartinBasti:
Is this woth it, given we have only one MD file and other are rst files?

ben-alkov:
Dropping, as we only have one MD file in here

PR #145, opened 2020-05-11
Option to allow OSBS autorebuilds to ignore isolated parent builds

rcerven:
* CLOUDBLD-926

### DIFF
--- a/docs/users.rst
+++ b/docs/users.rst
@@ -529,6 +529,9 @@ add_timestamp_to_release
   the container image will be available in container registry at:
   ``my-container-registry.example.com/fedora/rsyslog:32-5.20191007151825``.
 
+ignore_isolated_builds
+  Boolean to control whether to rebuild when parent image triggering build was isolated build.
+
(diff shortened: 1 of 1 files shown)


docs/users.rst
```diff
@@ -529,6 +529,9 @@ add_timestamp_to_release
   the container image will be available in container registry at:
   ``my-container-registry.example.com/fedora/rsyslog:32-5.20191007151825``.
 
+ignore_isolated_builds
+  Boolean to control whether to rebuild when parent image triggering build was isolated build.
```
athos-ribeiro, line 533: Could we have some context on this one? We could explain when this would be useful for a user

rcerven: well it is solely on user if he finds it useful, and I wouldn't like to make it more complicated as it is really simple option

MartinBasti:
can you disable stickler for this project?

rcerven:
> can you disable stickler for this project?

disabled

PR #149, opened 2020-05-19
Update Digests verification section

lkolacek:
CLOUDBLD-558

### DIFF
--- a/docs/users.rst
+++ b/docs/users.rst
@@ -974,6 +974,13 @@ in the manifest list do not match the analogous koji information.  This
 behavior can be deactivated through the ``deep_manifest_list_inspection``
 option. See `config.json`_ for further reference.
 
+Manifest lists can be manually pushed to the registry to make sure a specific tag
+(e.g., latest) is available for all platforms. In such cases, these manifest lists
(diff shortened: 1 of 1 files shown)


docs/users.rst
```diff
@@ -974,6 +974,11 @@ in the manifest list do not match the analogous koji information.  This
 behavior can be deactivated through the ``deep_manifest_list_inspection``
 option. See `config.json`_ for further reference.
 
+Manifest lists can be manually pushed to the registry to make sure a specific tag
+(e.g., latest) is available for all platforms. In such cases, these manifest lists
+may include images from different koji builds. We only want to check the digests
+for the images built in the current parent koji build.
```
athos-ribeiro, line 980: Wouldn't we want to check the digests for all the platforms requested for the *current* build?

athos-ribeiro: How about substituting

```
We only want to check the digests
for the images built in the current parent koji build.
```
for 

```
OSBS will only perform digest checks for the images requested in the current build. Moreover, build requests for platforms that were not built in the same koji build as the one found for the given image reference (manifest list) will fail.
```

lkolacek: Ack

PR #152, opened 2020-05-22
users: explain "modules" use-case, format, and behavior

ktdreyer:
Explain the use-case, expected format, and ODCS behavior for the "modules" option. In particular, we almost always want OSBS users to consume modules from Pulp rather than cherry-picking builds that might or might not be tested or signed.

### DIFF
--- a/docs/users.rst
+++ b/docs/users.rst
@@ -392,7 +392,30 @@ pulp_repos
   See :ref:`content_sets.yml`.
 
 modules
-  list of modules for requesting ODCS compose of type "module".
+  list of modules for requesting ODCS compose of type "module". ODCS will
+  cherry-pick each module into the compose.
+
+  Use this ``modules`` option to make module builds available that are not yet
+  available from the other options like Pulp. This is useful if you want to
+  test a newly-built module before it is available in Pulp, or if you want to
+  pin to a specific module that MBS has built.
+
+  This list can be of the format ``name:stream``, ``name:stream:version``, or
+  ``name:stream:version:context``.
+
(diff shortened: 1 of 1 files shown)


ktdreyer:
The corresponding ODCS documentation PR is https://pagure.io/odcs/pull-request/394

docs/users.rst
```diff
@@ -392,7 +392,30 @@ pulp_repos
...
+  If you specify a ``name:stream`` without specifying a ``version:context``,
+  ODCS will query MBS to find the very latest ``version:context`` build. For
+  example, if you specify ``go-toolset:rhel8``, ODCS will query MBS for the
+  latest ``go-toolset`` module build for the ``rhel8`` stream, whereas if you
+  specify ``go-toolset:rhel8:8020020200128163444:0ab52eed``, ODCS will compose
+  that exact module instead.
+
+  Note that if you simply specify a ``name:stream`` for a module, ODCS will
+  compose the very latest module that a module developer has built for that
+  stream, and this module may not be tested by QE or GPG signed.
```
ben-alkov, line 415: @ktdreyer; "may" is tricky[1] I think that "might" (specifically: "might not have been") probably better conveys your meaning.

[1] It can indicate "permission has been given"; "may not" can end up sounding like "QE isn't allowed to test this")

ktdreyer: Great suggestion. I've amended this text to use "might".

ben-alkov: Nice! Thanks.

gbenhaim:
LGTM

PR #155, opened 2020-06-03
Document 'from_latest' autorebuild feature

rcerven:
* CLOUDBLD-205


lkolacek:
@rcerven I agree with @athos we could modify `ignore_isolated_builds` and add default option ..  otherwise it seems to be clear what we already have.

rcerven:
> The descriptions provided are quite nice. However, this seems to be all documented a few lines above at https://github.com/containerbuildsystem/osbs-docs/blob/master/docs/users.rst#autorebuild. I'd either increment those descriptions or drop this PR.
> 
> I am not sure if these changes were not in place before the change was proposed or if @MartinBasti would like to see something else here.

I've spoke with @lcarva  and he said I should duplicate it so it is in config section and auto rebuilds section

athos-ribeiro:
> > The descriptions provided are quite nice. However, this seems to be all documented a few lines above at https://github.com/containerbuildsystem/osbs-docs/blob/master/docs/users.rst#autorebuild. I'd either increment those descriptions or drop this PR.
> > I am not sure if these changes were not in place before the change was proposed or if @MartinBasti would like to see something else here.
> 
> I've spoke with @lcarva and he said I should duplicate it so it is in config section and auto rebuilds section

Should we just add a link to that subsection instead of duplicating it then? Then it would be easier to maintain a single section on that topic.

MartinBasti:
When I reported issue, from_latest was not documented

ben-alkov:
> Should we just add a link to that subsection instead of duplicating it then

@athos-ribeiro, @rcerven; Yes

lcarva:
> > Should we just add a link to that subsection instead of duplicating it then
> 
> @athos-ribeiro, @rcerven; Yes

Yes, this is a good idea.

FWIW, this is not what I meant by duplicating them. I suggested duplicating this info in both the upstream and downstream docs, not on the same doc :)

rcerven:
> > > Should we just add a link to that subsection instead of duplicating it then
> > 
> > 
> > @athos-ribeiro, @rcerven; Yes
> 
> Yes, this is a good idea.
> 
> FWIW, this is not what I meant by duplicating them. I suggested duplicating this info in both the upstream and downstream docs, not on the same doc :)

ok will do

rcerven:
updated

PR #156, opened 2020-06-04
Deprecate additional-tags file

rcerven:
* CLOUDBLD-74


docs/users.rst
```diff
@@ -935,7 +935,8 @@ tagging the resulting image in the registry:
 
   - ``{version}`` (the ``version`` label)
   - ``latest``
-  - any additional tags named in the ``additional-tags`` file
+  - any additional tags named in the ``additional-tags`` file (this file is deprecated,
```
MartinBasti, line 938: Maybe make this more visible?

`....(DEPRECATED: use tags ....)`

rcerven: changed

rcerven:
> There is another entry in line 520. Also, consider being more explicit to say the feature will be removed in future versions

added that it will be removed in future versions,

line 520 is just saying that additional-tags will be ignored when you use tags in container.yaml (which is basically obsoleting that file anyway)

PR #159, opened 2020-06-26
Update cachito pkg_managers configuration behavior

athos-ribeiro:

### DIFF
--- a/docs/users.rst
+++ b/docs/users.rst
@@ -674,8 +674,13 @@ ref
 
 pkg_managers
   A list of package managers to be used for resolving the upstream project
-  dependencies. If not provided, cachito will try to figure out what manager to
-  use here.
+  dependencies. If not provided, Cachito will assume ``gomod`` due to backward
+  compatibility reasons, however, this default could be configured differently
+  on different Cachito deployments (make sure to check with your Cachito
+  instance admins). Finally, if this is set to an empty array (``[]``), Cachito
+  will provide the sources with no package manager magic. In other words, no
+  environment variables, dependencies, or extra configuration will be provided
+  with the sources.
 
 flags
   List of flags to pass to the cachito request. See the cachito_ documentation


athos-ribeiro:
Added a refference to a CLOUDBLD story and fixed backticks

docs/users.rst
```diff
@@ -674,8 +674,11 @@ ref
 
 pkg_managers
   A list of package managers to be used for resolving the upstream project
-  dependencies. If not provided, cachito will try to figure out what manager to
-  use here.
+  dependencies. If not provided, cachito will assume ``gomod`` due to backward
```
mprahl, line 677: Please keep in mind that the default package manager is deployment specific, so I'm not sure if that should be included in the upstream documentation.

athos-ribeiro: Doesn't Cachito set that as `gomod` for all envs by default? Having cachito not to set `gomod` as the value for DEFAULT_PACKAGE_MANAGERS is an opt-out feature and not opt-in, right?

Any suggestions on how to proper phrase this here? Moreover, do you think we should make this a required param in OSBS at some point?

I do feel that omitting this information could do more harm than good.

Finally, if we do not have it in place yet, it would be nice to throw a warning to let users know they should explicitly use the pkg_managers key.

mprahl: > Doesn't Cachito set that as `gomod` for all envs by default? Having cachito not to set `gomod` as the value for DEFAULT_PACKAGE_MANAGERS is an opt-out feature and not opt-in, right?

Yes, that is true.

> Any suggestions on how to proper phrase this here? Moreover, do you think we should make this a required param in OSBS at some point?
> 
> I do feel that omitting this information could do more harm than good.

How about after the end of the sentence, you add the following:
```
... compatibility reasons, however, this default could be configured differently on different Cachito deployments.
```
I like the idea of making it a requirement but it would involve us hunting down all getting existing Cachito users to have them change their `container.yaml` file, so it may not be worth the effort.

> 
> Finally, if we do not have it in place yet, it would be nice to throw a warning to let users know they should explicitly use the pkg_managers key.

athos-ribeiro: Fixed. I also made sure to let the user know they should check it with the cachito instance adimins.

mprahl: Looks good. The only thing that bugs me is that some places use `cachito` instead of `Cachito`.

docs/users.rst
```diff
@@ -674,8 +674,13 @@ ref
...
+  compatibility reasons, however, this default could be configured differently
+  on different Cachito deployments (make sure to check with your Cachito
+  instance admins). Finally, if this is set to an empty array (``[]``), cachito
+  will provide the sources with no package managers magic. In other words, no
```
mprahl, line 681:
```suggestion
  will provide the sources with no package manager magic. In other words, no
```

docs/users.rst
```diff
@@ -674,8 +674,13 @@ ref
...
+  environment variables or extra configuration will be provided with the
```
mprahl, line 682:
```suggestion
  environment variables, dependencies, or extra configuration will be provided with the
```

athos-ribeiro:
Comments addressed.

We should submit a different PR to s/cachito/Cachito/ in the rest of the documentation.

PR #161, opened 2020-06-30
Add doc of limiting image size

tkdchen:
* CLOUDBLD-1095

### DIFF
--- a/docs/admins.rst
+++ b/docs/admins.rst
@@ -250,6 +250,21 @@ For example, you might want to set up an HTTP proxy:
   - name: NO_PROXY
     value: localhost,127.0.0.1
 
+Limiting image size
+'''''''''''''''''''
+
+You can check the binary image's size before it is pushed to a registry. If it
+exceeds the configured size, the built image will not be pushed and the build
+fails.
+
+A typical configuration in reactor config map looks like::
+
+  image_size_limit:
+    binary_image: 10000
+
+The value is the size in bytes of uncompressed layers. When either
+``binary_image`` or ``image_size_limit`` is omitted, or if ``binary_image`` is
+set to ``0``, the check will be skipped.
 
 Setting up koji for container image builds
 ------------------------------------------


docs/admins.rst
```diff
@@ -250,6 +250,20 @@ For example, you might want to set up an HTTP proxy:
...
+Limiting image size
+'''''''''''''''''''
+
+It is doable to limit the binary image size before pushing to registry. If a
+binary image's size exceeds configured image size, it will be refused to be
+pushed. A typical configuration looks like::
+
+  "image_size_limit": {
+    "binary_image": 10000
+  }
```
athos-ribeiro, line 262: Nitpick: We could drop the brackets to match the other  reactor configmap examples in the docs. Moreover, when the top level section (where this subsection is contained) starts, it does not explicit mentions these entries are to be contained in the reactor configmap. We could make it a bit more explicit.

Not a blocker though.

docs/admins.rst
```diff
@@ -250,6 +250,20 @@ For example, you might want to set up an HTTP proxy:
...
+
+The value is the size in bytes of uncompressed layers. When either
+``binary_image`` or ``image_size_limit`` is omitted, or setting ``0`` to
+``binary_image``, the check will be skipped.
```
ben-alkov, line 266: @tkdchen; Sorry, I don't know ho to do a block diff in GFMD (not sure it's even possible...). Suggestion below:

```
You can check the binary image's size before it is pushed to a
registry. If it exceeds the configured size, the push will fail.

A typical configuration looks like::

  "image_size_limit": {
    "binary_image": 10000
  }

The value is the size in bytes of the uncompressed layers. When
either ``binary_image`` or ``image_size_limit`` is omitted, or
if ``binary_image`` is set to ``0``, the check will be skipped.
```

tkdchen:
@athos-ribeiro @ben-alkov Thanks for your thought and suggestion. PTAL.

PR #164, opened 2020-07-08
Document Image Content Manifests creation

athos-ribeiro:
* CLOUDBLD-1117

### DIFF
--- a/docs/users.rst
+++ b/docs/users.rst
@@ -785,21 +785,6 @@ An example::
   - server-for-power-le-rpms
   - server-extras-for-power-le-rpms
 
-OSBS will create a /root/buildinfo/content_manifests/`{IMAGE_NVR}.json` file in
-the final built image containing platform specific content sets information.
-
-For instance::
-
-    {
-      "metadata": {
-        "image_layer_index": 3
-      },
-      "content_sets": ["server-rpms", "server-extras-rpms"]
-    }
-
-where `image_layer_index` is the index for the most recent layer for that
-image.
-
 Using Artifacts from Koji
 -------------------------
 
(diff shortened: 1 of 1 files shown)


docs/users.rst
```diff
@@ -1618,6 +1603,66 @@ Manifests will not be pushed to the application registry for scratch builds,
...
+information on the contents shipped in the image.
+
+The Image Content Manifest file is layer specific, and is located under the
+``/root/buildinfo/content_manifests`` directory. It is named after the image NVR,
+and it is validated against the JSON Schema that defines the `Image Content Manifest`_.
+
+
+Among the data available in the Image Content Manifest file (Check the JSON
+Schema for further information), most important are the ``image_layer_index``,
+which point to the layer that introduced the components listed in that file,
```
rcerven, line 1642: can we be more specific about, like we had before?:
`image_layer_index` is the index for the most recent layer for that image

docs/users.rst
```diff
@@ -1618,6 +1603,66 @@ Manifests will not be pushed to the application registry for scratch builds,
...
+and:
+
+.. _`Image Content Manifest`: https://github.com/containerbuildsystem/atomic-reactor/blob/master/atomic_reactor/schemas/content_manifest.json
+
+Content Sets
+~~~~~~~~~~~~
+
+The ``content_sets`` field lists the content sets used during the image build
```
rcerven, line 1650: "the content sets used during the image build"

this isn't exactly truth

because they can have content_sets.yaml BUT not have container.yaml
in that case content sets weren't used, but will be listed here

docs/users.rst
athos-ribeiro: done

docs/users.rst
athos-ribeiro: Changed to say it is whatever the user has specified in dist-git for the given platform

chmeliik: Should we specify dist-git in upstream docs? I found it in a few other places in the docs as well, but isn't it quite specific to our use case?

athos-ribeiro: Good question. I am not sure how well koji deals with non dist-git repositories nowadays (I found only a single reference to the lookaside cache in the code). OSBS upstream does reference it quite often though.

I do agree it is not needed to say dist-git here though. Changed to `git`

PR #166, opened 2020-08-10
Build only content sets used just for building content

rcerven:
* CLOUDBLD-2046


docs/users.rst
```diff
@@ -461,6 +462,13 @@ modular_koji_tags
   When ``true`` is specified instead of list, koji build tag of the koji build target
   will be used instead.
 
+build_only_content_sets
+  Content sets used only for building content, not for distributing.
+  Will be used only if ``pulp_repos`` is set to true. These content sets
+  won't be included in ICM.
```
MartinBasti, line 468: There is no explanation what ICM is (not even in this doc)

rcerven: added link

PR #175, opened 2021-01-29
users: improve content set examples

ktdreyer:
Rephrase the sentence about Pulp repository IDs to use the active voice.

Show how users can use content set names or repository IDs with RHEL 7 and RHEL 8 examples that are easy to copy and paste.


ben-alkov (approved):

tkdchen (approved):
Thank you very much.

PR #178, opened 2021-03-09
Allow setting custom self-signed cert for ODCS

tkdchen:
* CLOUDBLD-4345

### DIFF
--- a/docs/admins.rst
+++ b/docs/admins.rst
@@ -266,6 +266,23 @@ The value is the size in bytes of uncompressed layers. When either
 ``binary_image`` or ``image_size_limit`` is omitted, or if ``binary_image`` is
 set to ``0``, the check will be skipped.
 
+Custom CA bundle
+''''''''''''''''
+
+It is allowed to specify a custom CA bundle explicitly to include self-signed
+certificates. If set, it will be injected into every YUM repository added by
+users. The custom CA bundle is used during the container build process only.
+
+Set the CA bundle certificate by config ``builder_ca_bundle`` at the top level
+of the reactor config. The value must be a file name with an absolute path to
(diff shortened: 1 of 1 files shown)


docs/admins.rst
```diff
@@ -229,6 +229,13 @@ default_signing_intent
   Name of the default signing intent to be used when one is not provided
   in ``container.yaml``.
 
+To allow accessing repositories generated by ODCS via HTTPS, set
```
MartinBasti, line 232: That's not true, you need this to allow custom selfsigned CA certificates. It depends what kind of cert is ODCS using

docs/admins.rst
```diff
@@ -229,6 +229,15 @@ default_signing_intent
   Name of the default signing intent to be used when one is not provided
   in ``container.yaml``.
 
+If HTTPS is enabled for the access to YUM repositories hosted in ODCS,
```
MartinBasti, line 232: Not just ODCS, we have implemented it for any yum repo, even those passed by users via parameter.

tkdchen: Well, if so, this is not the right section to add the new config description. This section is dedicated for odcs. Which section is proper to have a general description for the `builder_ca_bundle`?

MartinBasti: Just make it as a new section see `Limiting image size` for example

MartinBasti: It can be section `Custom CA bundle for yum repos` or so

tkdchen:
@MartinBasti PTAL.

MartinBasti (approved):
Thank you!

mkosiarc (approved):

lkolacek (approved):

rcerven (approved):

PR #182, opened 2021-03-30
users: explain layered scratch builds

ktdreyer:
Scratch builds can use other scratch builds in `FROM` instructions. Explain how this works and give an example.

Note, I've not tested this behavior myself, yet. I'm just capturing a discussion with @lcarva here.


docs/users.rst
```diff
@@ -1092,6 +1092,13 @@ that was pushed to ``rsyslog`` will cause a build failure.
 
 This behavior requires koji integration to be enabled in the OSBS environment.
 
+For scratch builds, OSBS does not resolve a Dockerfile ``FROM`` instruction
```
rcerven, line 1095: what do you mean by the 1st sentence???

lcarva: I think it's meant to capture that OSBS will not ensure there's a corresponding Koji build, i.e. not run koji_parent plugin.
I don't know how to rephrase it to make it easier to understand.

rcerven (approved):

MartinBasti (approved):

MartinBasti:
Have you tested it?

PR #214, opened 2022-04-21
Document valid pattern for floating tags

gsaslis:
I came across this issue when trying to use a tag that contains a `-` in an OpenShift Build Service (OSBS) build for 3scale API Management.

### DIFF
--- a/docs/users.rst
+++ b/docs/users.rst
@@ -1202,7 +1202,8 @@ tagging the resulting image in the registry:
   over time. In other words, every time you build a new container image,
   OSBS updates these floating tags. Examples: ``latest``, or ``{version}``
 
-  The floating tags must match the pattern `^[\w.]{0,127}$`. 
+  The floating tags must match the pattern `^[\w.]{0,127}$`. For example, 
+  `3scale2.12.dev` is a valid tag, but `3scale2.12-dev` isn't.
   
   Floating tags are configurable. If you set ``tags`` in container.yaml, OSBS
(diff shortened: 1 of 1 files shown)


docs/users.rst
```diff
@@ -1202,6 +1202,8 @@ tagging the resulting image in the registry:
   over time. In other words, every time you build a new container image,
   OSBS updates these floating tags. Examples: ``latest``, or ``{version}``
 
+  The floating tags must match the pattern `^[\w.]{0,127}$`. 
+
```
tkdchen, line 1206: A concrete example is always better than giving such a regular expression in documentation. We can't assume a user is familiar or even master the regular expression, and this regex is not much straightforward for a user to have a quick understand what a floating tag could look like. The unique tag example mentioned above is a good example to follow I think.

gsaslis: While I don't think this should block merging this PR (as it already improves the situation - where the docs provide neither a concrete example, nor the validation rule), I'll go ahead and add one. ;)

gsaslis: @tkdchen resolving this after adding an example, please reopen if you need something further.

tkdchen:
@gsaslis You have to sign-off your commits.

MartinBasti:
Any update on this PR?

gsaslis:
@tkdchen / @MartinBasti please take a look.

MartinBasti (approved):

ben-alkov (approved):

PR #217, opened 2022-06-20
Update reactor config map example with new registries_cfg_path config

tkdchen:
* CLOUDBLD-9346

### DIFF
--- a/docs/build_parameters.rst
+++ b/docs/build_parameters.rst
@@ -34,7 +34,9 @@ The environment configuration is supplied as a ``ConfigMap`` like::
 The schema definition `config.json`_ in atomic-reactor contains a description
 for each property.
 
-Example of **REACTOR_CONFIG**::
+Example of **REACTOR_CONFIG**:
+
+.. code-block:: yaml
 
     version: 1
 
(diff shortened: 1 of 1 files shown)


chmeliik (commented):
Could we also explain what format the file should have (this one, I believe? https://www.mankier.com/5/containers-auth.json) and that it's the auth file for *all* push and pull registries?

tkdchen:
The whole "Reactor Configuration" section looks incomplete. There is still confusion about the configuration file. For example,

* `Dockercfg` accepts both file names `.dockercfg` and `.dockerconfigjson`, and both of these two kinds of configuration accept two formats, one has `auths` key but another one doesn't. The comment tells "# If the auths key exist then we have a dockerconfigjson secret.", however, both the link you gave and an AWS doc I found tell the `auths` key works for `.dockercfg` as well.
* Is the configuration file used by a-r code only? How the different format affects the podman-remote by passing via the `--authfile`?

A few things are still not clear enough yet. I think it would be good to explain the relative part of reactor config in a dedicated story. And perhaps some "dead code" can also be removed, e.g. `unpack_auth_b64`.

chmeliik:
+1 for dedicated story :+1:

chmeliik (approved):
LGTM :+1:

Issue #163, opened 2020-07-07
"go" settings are deprecated


ktdreyer:
We should update the documentation to steer users away from using the `go` key in `container.yaml`. This is the old deprecated way of performing container-first builds in OSBS.

Cachito is the replacement. To switch to Cachito, set the `remote_source` key instead. OSBS does not permit users to specify a `go` key with a `remote_source` key.

PR #219, opened 2022-06-27
Document deprecated "go" settings

sumincho22:
CLOUDBLD-8399

Signed-off-by: Sumin Cho <sucho@redhat.com>

Reference to Issue #163:

> We should update the documentation to steer users away from using the go key in container.yaml. This is the old deprecated way of performing container-first builds in OSBS. Cachito is the replacement. To switch to Cachito, set the remote_source key instead. OSBS does not permit users to specify a go key with a remote_source key.

I have added this information to the warning tag under "go" section.

### DIFF
--- a/docs/users.rst
+++ b/docs/users.rst
@@ -390,7 +390,7 @@ not
 go
 ~~
 
-.. warning:: Using this key is deprecated in favor of using Cachito integration
+.. warning:: Using this key is deprecated in favor of using Cachito integration. To switch to Cachito, set the ``remote_sources`` key instead. OSBS does not permit users to specify a ``go`` key with a ``remote_sources`` key.
 
 Keys in this map relate to source code in the Go language which the
 user intends to be built into the container image. They are


docs/users.rst
```diff
@@ -390,7 +390,7 @@ not
 go
 ~~
 
-.. warning:: Using this key is deprecated in favor of using Cachito integration
+.. warning:: Using this key is deprecated in favor of using Cachito integration. To switch to Cachito, set the ``remote_source`` key instead. OSBS does not permit users to specify a ``go`` key with a ``remote_source`` key.
```
MartinBasti, line 393: IMO we should use `remote_sources` (plural) https://osbs.readthedocs.io/en/latest/users.html#remote-sources, singular `remote_source` will be deprecated too

sumincho22: Gotcha, committed this change.

mkosiarc (approved):

MartinBasti (approved):

PR #220, opened 2022-08-15
undocument cachito single remote source

taylormadore:
CLOUDBLD-6771

Since using multiple remote sources in cachito is
preferred and because we intend to deprecate the
single remote source option in the future, multiple
remote sources should be presented as the only option

Signed-off-by: Taylor Madore <tmadore@redhat.com>

### DIFF
--- a/docs/admins.rst
+++ b/docs/admins.rst
@@ -494,6 +494,8 @@ This section describes how to configure OSBS to use cachito as described above.
 :ref:`cachito-usage` describes how to get OSBS to use cachito in
 a specific container build, as an OSBS user.
 
+.. _configure-cachito-instance:
+
 Configuring your cachito instance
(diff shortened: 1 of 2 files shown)


rcerven:
it shouldn't be removed, but marked as deprecated

taylormadore:
I think Martin suggested in the story that we remove single remote source from the docs entirely because it's not the preferred way anymore and it confuses users having both in the docs. @MartinBasti what do you think?

MartinBasti:
yes, it puzzles users a lot, they always mess plural and singular form.
We should introduce user warnings later too, once initial OSBS2 is done

lkolacek (approved):
LGTM :+1:

rcerven (approved):

ejegrova (approved):
LGTM

PR #228, opened 2022-11-22
add documentation for configurable build time

hjmodi:
* CLOUDBLD-9913

Signed-off-by: Harsh Modi <hmodi@redhat.com>

### DIFF
--- a/docs/admins.rst
+++ b/docs/admins.rst
@@ -51,6 +51,21 @@ Specifies the cleanup strategy for OSBS builds. By default this is set to
 ``true``. When it is set to ``true`` it'll remove the build pipeline on a
 completed or failed build.
 
+default_buildtime_limit
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Specifies default build time limit. If users don't specify a build time
+limit, this limit is used. It is integer set in seconds. If this value
+is not specified, the default value set by maintainers is used.
+
+max_buildtime_limit
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Specifies max build time limit. This is used to cap build time limit
+set by the user. If the user specifies build time limit greater than
+this value an exception will be thrown. If this value is not specified,
+the default value set by maintainers is used.
+
 Deploy OSBS on OpenShift
 ------------------------
 
--- a/docs/users.rst
+++ b/docs/users.rst
@@ -412,6 +412,16 @@ modules
   path
     path to directory containing source code (or its parent), possibly within archive.
 
+.. _container.yaml-buildtime_limit:
+
+buildtime_limit
+~~~~~~~~~~~~~~~
+
+This parameter is used for setting a build time limit in seconds. After specified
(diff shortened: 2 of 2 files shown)


docs/admins.rst
```diff
@@ -51,6 +51,21 @@ Specifies the cleanup strategy for OSBS builds. By default this is set to
 ``true``. When it is set to ``true`` it'll remove the build pipeline on a
 completed or failed build.
 
+default_buildtime_limit
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Specifies default build time limit. If users don't specify a build time
+limit, this limit is used. It is integer set in seconds. If this value
+is not specified, the default value of ``10800`` is used.
```
rcerven, line 59: I'd rather omit mentioning specific value

docs/admins.rst
```diff
@@ -51,6 +51,21 @@ Specifies the cleanup strategy for OSBS builds. By default this is set to
...
+
+max_buildtime_limit
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Specifies max build time limit. This is used to cap build time limit
+set by the user. If the user specifies build time limit greater than
+this value an exception will be thrown. If this value is not specified,
+the default value of ``21600`` is used.
```
rcerven, line 67: I'd rather omit mentioning specific value

docs/admins.rst
hjmodi: removed specific value

docs/admins.rst
hjmodi: removed specific value

rcerven (approved):

docs/admins.rst
```diff
@@ -51,6 +51,21 @@ Specifies the cleanup strategy for OSBS builds. By default this is set to
 ``true``. When it is set to ``true`` it'll remove the build pipeline on a
 completed or failed build.
 
+default_buildtime_limit
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Specifies default build time limit. If users don't specify a build time
+limit, this limit is used. It is integer set in seconds. If this value
+is not specified, the default value set by maintainers is used.
```
mkosiarc, line 59: Maybe I am misunderstanding, but the values default_buildtime_limit and max_buildtime_limit have to be always set by admins and not users right? If that's the case then the following line does not make sense to me:

> If this value is not specified, the default value set by maintainers is used.

hjmodi: I am referring to `buildtime_limit` that can be set in `container.yaml`

mkosiarc: I see now. I was just a bit confused because in sentence "this limit is used" *this* refers to default_buildtime_limit but in "If this value is not specified" *this* refers to buildtime_limit.

mkosiarc (approved):

docs/admins.rst
mkosiarc: Ideally, if would like if this was a bit clearer (for people like me:D) but I acked this.

PR #233, opened 2023-04-06
users: clarify "compose" section operation

ktdreyer:
Clarify that a `compose` section simply makes the `.repo` file available in the build environment. It does not install any packages into the image.


chmeliik (approved):

lkolacek (approved):
Looks great! :+1: Thank you

